Issue #4, opened 2021-08-14
The key abstraction of pipeline framework and component interface design 


GuoRentong:
# A Hello World Example
---
```python
class Inc(Operator):
    @create_op_in_pipeline
    def __call__(self, x:int) -> int:
        return x+1

class Add(Operator):
    @create_op_in_pipeline
    def __call__(self, x1:int, x2:int) -> int:
        return x1+x2

@create_pipeline
def my_pipeline(x1:int, x2:int):
    inc1 = Inc()
    inc2 = Inc()
    add = Add()
    return add(inc1(x1), inc2(x2))
```

In this example, we have three operators in a pipeline, where *add* depends on the results of *inc1* and *inc2*. Behind the scene, Towhee's compiler will create a DAG during the execution of *my_pipeline*. The DAG construction is driven by the operator decorator and pipeline decorator. When the program executes the line *add(inc1(x1), inc2(x2))*,  the decorator @create_op_in_pipeline will be called before the Operators' __call__ method. It will put *inc1*, *inc2*, *add* into the pipeline's context, link the pipeline's input *x1*, *x2* to *inc1*, *inc2* respectively, and set *inc1*, *inc2*'s output as *add*'s input. The operators' dependencies are also settled during the DAG construction.

# Main Components
---
To support the example mentioned above, Towhee needs five major components.

* **Operator** An operator is a set of code that performs one step in the pipeline, such as preprocessing, model inference, postprocessing, etc. An operator is analogous to a function, in that it has a name, parameters, return values, and a body. Once involved in a pipeline, an operator will be construct as a node in the DAG.

* **Compiler** There are two phase compilation. The first step is to convert a pipeline description (eg. *my_pipeline*) and operators (eg. *Inc, Add*) into a intermediate graph, where the variables, operators, dependencies are precisely described. The second step is to convert the intermediate graph to a backend executable, such as a local python-driven DAG, or a kubeflow pipeline.

* **Variable** The data abstraction of pipeline, operator's inputs and outputs. It is partitionable and iterable. The difference of device memory, the cross-device memory copy, data transfer is also handled by Variable.

* **Pipeline** The runtime scheduling context of a pipeline, including variables, tasks, and execution flow. All the statefull part during taks execution is maintained in *Pipeline*.

* **Engine** The engine driving the task execution on a pipeline. It takes a *Pipeline* as its scheduling context, and performs all the necessary data paritition, operator parallelization, resource management, etc.

# Works Remaining
---
#### Operator
[...]

PR #16, opened 2021-08-25
interface of dataframe

GuoRentong:
* interface of dataframe
* interface of dataframe iterator
* interface of dataframe representation in DAG


GuoRentong:
#4

PR #21, opened 2021-08-27
Add basic interface of Task, TaskQueue, TaskExecutor

GuoRentong:
In the current design, we bind one TaskQueue and one Task Executor to each device.
A Task Executor may have multiple Operators to handdle.


junjiejiangjjj (approved):
/approve
/lgtm

PR #22, opened 2021-08-27
add trainer and its utilities

oneseer:
add vision transformer and its utilities


GuoRentong (changes requested):
Too many things in a single PR.
We'd better create PR separately for each individual trainer utility

PR #24, opened 2021-08-28
cnn trainer design

derekdqc:
design the cnn trainer.


towhee/trainer/cnn_trainer.py
```diff
@@ -0,0 +1,629 @@
...
+    import datasets
+
+
+if TYPE_CHECKING:
+    import optuna
+
+logger = logging.get_logger(__name__)
+
+
+class Trainer:
```
GuoRentong, line 140: should we consider a more specific Trainer name such as TorchCNNTrainer, since we may have other types of Trainer. 
In addition, do we need a base class 'TorchTrainer' for those basic methods in Pytorch training?

GuoRentong (changes requested):

towhee/data/image_datasets.py
```diff
@@ -0,0 +1,55 @@
+import os
```
GuoRentong, line 1: add lisence for this file

towhee/data/image_datasets.py
```diff
@@ -0,0 +1,55 @@
...
+
+import pandas as pd
+from pandas import Series
+from torch.utils.data import Dataset
+from PIL import Image
+
+
+class ImageDataset(Dataset):
+    """
+        PytorchImageDataset is a dataset class for PyTorch.
```
GuoRentong, line 11: TorchImageDataset or ImageDataset?
There are two different names in class declaration and docstring

GuoRentong (changes requested):

PR #36, opened 2021-09-01
cnn trainer and dataset

derekdqc:
cnn trainer and dataset design


fzliu:
Can we add a simple README or example script that utilizes the trainer?

PR #43, opened 2021-09-02
Add abstraction of engine and scheduler

GuoRentong:
* abstraction of engine
* abstraction of scheduler
* an example of FIFO scheduler


junjiejiangjjj (approved):
/approve
/lgtm

PR #54, opened 2021-09-06
image dataset implementation for training.

derekdqc:


GuoRentong (changes requested):
please add unittest for PyTorchImageDataset

PR #57, opened 2021-09-06
format code

junjiejiangjjj:
Use pylint format code


fzliu:
Does the entire codebase now pass all pylint checks?

Issue #58, opened 2021-09-06
Engine implementation, testing, and my-first-pipeline


fzliu:
# Engine Implementation Checklist

The engine lies at Towhee's core, driving the dataflow, partitioning tasks to devices, and managing the overall state of the pipeline. When a user calls a `pipeline()` function, an `Pipeline` object and all associated contexts and device managers are created. Execution threads for each device is created through a `TaskExecutor`, and management of graph state is done through `GraphContext` and `OperatorContext` classes. Each `OperatorContext` is responsible for passively managing its own input iterators, while the `Scheduler` actively (periodically) asks each operator whether or not it has any tasks ready for processing. Each task is then assigned by the `Scheduler` to one of many `TaskExecutor` objects, resulting in parallel processing of simultaneous tasks.

This architecture allows a single pipeline to be run across one or many interconnected machines, each of which may have multiple devices.

# Task List
---

#### `GraphContext` and `OperatorContext` (~19 days)
- [x] graph state management
- [x] operator state management
- [x] data management and dataflow between operators
- [x] task acquisition, if input data is available
- [ ] unit tests

#### `Dataframe` and `Array` (~3 days)
- [x] integration with data source
- [x] `map` iterator (v0.1)
- [x] unit tests

#### `Task`, `TaskExecutor`, and `OperatorPool` (~9 days)
- [x] operator acquisition & release
- [x] task management
- [x] task execution
- [x] unit tests

#### `Scheduler` (~5 days)
- [x] looping FIFO scheduler implementation
- [x] basic integration tests with simple Operators
- [x] unit tests

#### `Pipeline` (~2 days)
- [x] `GraphContext` wrapper implementation
- [ ] unit tests

#### my-first-pipeline (~3 days)
- [ ] successful creation of pipeline using DAG
- [ ] successful processing with PyTorch models

PR #60, opened 2021-09-07
Task, TaskExecutor, OperatorPool implementation

fzliu:
See section 3 of task list for #58.


towhee/engine/operator_loader.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # Lookup the path for the operator in local cache. Example directory structure:
+        #  /home/user/.towhee/cache/organization-name/operator-name
+        #   |_ /home/user/.towhee/cache/organization-name/operator-name/operator.py
+        #   |_ /home/user/.towhee/cache/organization-name/operator-name/resnet.torch
+        #   |_ /home/user/.towhee/cache/organization-name/operator-name/config.json
+        paths = list((self._cache_path / function).glob('[!_]*.py'))
+        if not paths:
+            raise TypeError('Operator definition not found')
+        if len(paths) > 1:
```
junjiejiangjjj, line 60: An operator can have many *.py files, we just use operator.py to create an operator object.

towhee/engine/operator_pool.py
```diff
@@ -13,23 +13,56 @@
...
+
+        Returns:
+            (`towhee.OperatorBase`)
+                The operator instance reserved for the caller.
         """
-        Release an Operator.
+        if task.op_name not in self._all_ops:
+            op = self._op_loader.load_operator(task.op_func, task.op_args)
+            op.name = task.op_name
+        else:
```
junjiejiangjjj, line 51: Use op_name and init params to calc a key, because the same operator with different init params is a different object.

fzliu: Good catch, I'll fix this.

towhee/engine/operator_loader.py
fzliu: I initially had it just named as `operator.py`, but during testing phase it conflicted with the `operator` standard library. Perhaps just `definition.py` or `operator_def.py`?

towhee/engine/task_executor.py
```diff
@@ -26,51 +27,59 @@ class TaskExecutor(threading.Thread):
...
+        Args:
+            task:
+                Pre-initialized `Task` object to push onto the queue.
         """
-        # todo self._task_queue.push(task)
-        raise NotImplementedError
+        return self._task_queue.push(task)
 
     def run(self):
-        while not self._need_stop:
```
junjiejiangjjj, line 66: One task_executor should have multi work thread to run tasks

fzliu: I'll add a TODO for this.

PR #62, opened 2021-09-07
Add operator base

junjiejiangjjj:
1. OperatorBase class
2. Implementation a simple operator: AddOperator

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,10 +16,13 @@
 from setuptools import setup, find_packages
 
 import unittest
+
+
 def test_suite():
     test_loader = unittest.TestLoader()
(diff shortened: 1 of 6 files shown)


fzliu:
What is the difference between `Operator` and `OperatorBase` (they are both ABCs)? Isn't `Operator` supposed to serve as the base class that other operators inherit from?

junjiejiangjjj:
> What is the difference between `Operator` and `OperatorBase` (they are both ABCs)? Isn't `Operator` supposed to serve as the base class that other operators inherit from?

`Operator` was previously designed. Now we have a new high-level interface, so I Redesigned it: `OperatorBase`

Issue #56, opened 2021-09-06
DAG implementation, testing, and integration with engine components


fzliu:
# DAG Implementation Checklist

Based on the agreed upon architecture, the DAG will serve as a template for the way the engine moves data around in the pipeline. In essence, each engine component has a corresponding representation in the DAG:

* `GraphRepr`: A simple container for operators and dataframes. It does not explicitly hold any data regarding the dataflow between operators and dataframes.
* `OperatorRepr`: An operator abstraction. This will hold the operator function along with input and output dataframes and the iterators used for the respective dataframes.
* `DataframeRepr`: A dataframe representation. Essentially a collection of variable representations with no information about the operators which use or supply said dataframes.
* `VariableRepr`: A variable representation. Stores information about the individual variable, i.e. variable type (`ndarray`, `Image`, etc) without knowing which Dataframe it is a part of.

# Task List
---
#### Graph
- [ ] base implementation (attributes & helper functions)
- [ ] unit tests
- [ ] `from_yaml`
- [ ] `to_yaml`
- [ ] error checking: no cyclical loops

#### Operator
- [ ] base implementation (attributes & helper functions)
- [ ] unit tests
- [ ] error checking: no orphan operators
- [ ] error checking: all operators have both inputs and outputs

#### Dataframe
- [ ] base implementation (attributes & helper functions)
- [ ] unit tests
- [ ] `from_input_annotations`
- [ ] `from_output_annotations`
- [ ] error checking: no orphan dataframes
- [ ] error checking: all dataframes have input operators (except data sources)

#### Variable
- [ ] base implementation (attributes & helper functions)
- [ ] unit tests

#### Other Tests
- [ ] load yaml -> dump yaml identity checking tests
- [ ] engine can create pipeline using graph

PR #69, opened 2021-09-10
Implement components representations in dag

Chiiizzzy:
#56 

Implemented a Graph representation as the following image:
GraphRepr


towhee/base_repr.py
```diff
@@ -25,30 +22,32 @@ class BaseRepr:
...
-            value = self.__getattribute__(name)
-            if name[0] != '_' and not isinstance(value, types.FunctionType):
-                value = value.__class__.__name__
-                out += name + ': ' + str(value) + '\n'
-        return out
+    @name.setter
+    def name(self, value):
+        self._name = value
+
+    # def serialize(self) -> str:
```
fzliu, line 37: This implementation of `serialize` is no longer correct based on our latest arch, remove this commented block and add a `raise NotImplementedError` instead.

towhee/dag/graph_repr.py
```diff
@@ -21,16 +23,25 @@
...
-    are used during execution to load functions and pass data to the correct operators.
-    """
+    """This class presents a complete representation of a graph.
 
-    def __init__(self, name: str):
-        # TODO(Chiiizzzy)
+    A graph contains individual subcomponents, including Operators, Dataframes, and
+    Variables. Graph representations are used during execution to load functions and
+    pass data to the correct operators.
+    """
```
fzliu, line 31: Add `__init__` argument descriptions in the class docstring here.

towhee/dag/graph_repr.py
```diff
@@ -21,16 +23,25 @@
...
         self._operators = {}
         self._dataframes = {}
+        # if `file` is a loacl file
+        if file_or_url and os.path.isfile(file_or_url):
+            with open(file_or_url, 'r', encoding='utf-8') as f:
+                self._src = yaml.safe_load(f)
+            self.from_yaml(self._src)
+        # if `file` is a remote url
+        elif file_or_url:
+            self._src = requests.get(file_or_url).content.decode('utf-8')
```
fzliu, line 43: Let's add a test case for remote file upload as well. As a source, we can create a gist for `towhee-io`.

towhee/dag/graph_repr.py
```diff
@@ -40,15 +51,33 @@ def operators(self) -> Dict[str, OperatorRepr]:
...
+            raise ValueError('src is not a valid YAML file')
+        if not essentials.issubset(set(graph_dict.keys())):
+            raise ValueError('src cannot descirbe a DAG in Towhee')
+
+        # load name from YAML
+        self._name = graph_dict['graph']['name']
+
+        # load dataframes from YAML
+        for df in graph_dict['dataframes']:
+            self._dataframes[df['name']] = DataframeRepr(df['name'], str(df))
```
fzliu, line 76: Are you sure we can use `str(df)` to transform this back into a YAML string? Isn't `df` a nested dictionary?

towhee/base_repr.py
Chiiizzzy: fixed

towhee/dag/graph_repr.py
Chiiizzzy: fixed

towhee/dag/graph_repr.py
Chiiizzzy: My bad, it loads dataframe from a string representation of a dict. Still, yaml.safe_load can load a dict from it

fzliu: Weird, are you sure the resulting representation is still correct in all cases?

Chiiizzzy: I've checked again and printed the result. It seems fine. In which case do you think will go wrong?

fzliu: I am not 100% sure, but the output of `str(df)` is not YAML, there are probably cases where it will not work properly

Chiiizzzy: fixed

towhee/base_repr.py
```diff
@@ -25,30 +22,26 @@ class BaseRepr:
         name:
             Name of the internal object described by this representation.
     """
-
-    def __init__(self, name: str):
-        self._name = name
+    def __init__(self, name: str = None):
+        if name:
```
junjiejiangjjj, line 26: if the name is None, just set self._name = None，make sure this attr is exist

towhee/base_repr.py
```diff
@@ -25,30 +22,26 @@ class BaseRepr:
...
+            self._name = name
 
     @property
     def name(self):
         return self._name
 
+    @name.setter
+    def name(self, value):
```
junjiejiangjjj, line 34: value: str

towhee/dag/graph_repr.py
```diff
@@ -21,16 +23,31 @@
...
+    Variables. Graph representations are used during execution to load functions and
+    pass data to the correct operators.
+
+    Args:
+        name(`str`):
+            The representation name.
+        file_or_url(`str`):
+            The file or remote url that stores the information of this representation.
+    """
+    def __init__(self, name: str, file_or_url: str = None):
```
junjiejiangjjj, line 38: We may support local file/HTTP/HDFs or other file systems, so we can use file_str, check types, 
HTTP:  file_str.low().startwiths("http")
HDFS: file_str.low().startwiths("hdfs")
local: file_str.low().startwiths('/') and so on. Write a function to check  the type and read the  file

towhee/dag/graph_repr.py
```diff
@@ -40,15 +57,33 @@ def operators(self) -> Dict[str, OperatorRepr]:
     def dataframes(self) -> Dict[str, DataframeRepr]:
         return self._dataframes
 
-    def from_yaml(self, yaml: str):
+    def from_yaml(self, src: str):
```
junjiejiangjjj, line 60: Use staticmethod, and return repr object

towhee/dag/operator_repr.py
```diff
@@ -30,26 +29,53 @@ class OperatorRepr(BaseRepr):
...
+    def inputs(self, value: Dict[str, DataframeRepr]):
         self._df_in = value
 
-    @df_out.setter
-    def df_out(self, value: DataframeRepr):
+    @outputs.setter
+    def outputs(self, value: Dict[str, DataframeRepr]):
         self._df_out = value
+
+    def from_yaml(self, src: str, dataframes: Dict[str, DataframeRepr]):
```
junjiejiangjjj, line 60: staticmethod, return a repr, can use another function to add  DatafremeRepr object

towhee/dag/variable_repr.py
```diff
@@ -21,17 +20,16 @@ class VariableRepr(BaseRepr):
...
         dtype:
             A string or instance of `numpy.dtype` indicating the internal data type for
             this variable.
     """
-
-    def __init__(self, name: str, vtype: str, dtype: str):
-        super().__init__(name)
+    def __init__(self, vtype: str, dtype: str):
+        super().__init__()
         self._vtype = str(vtype)
```
junjiejiangjjj, line 33: Input is already str type, why use str(...)

towhee/base_repr.py
Chiiizzzy: fixed

towhee/base_repr.py
Chiiizzzy: fixed

towhee/dag/graph_repr.py
Chiiizzzy: changed

towhee/dag/variable_repr.py
Chiiizzzy: fixed

towhee/dag/operator_repr.py
Chiiizzzy: fixed

towhee/dag/graph_repr.py
Chiiizzzy: fixed

towhee/dag/dataframe_repr.py
```diff
@@ -16,79 +16,73 @@
...
-    def __setitem__(self, key: str, value: VariableRepr):
-        """Sets a single variable representation within the dataframe.
+        self._columns = []
+
+    @property
+    def columns(self):
+        return self._columns
+
+    @columns.setter
+    def columns(self, value: VariableRepr, index: int):
```
junjiejiangjjj, line 49: Tow function named colums

towhee/dag/dataframe_repr.py
```diff
@@ -16,79 +16,73 @@
...
-            was provided.
-        """
-        return self._columns.get(key)
-
-    def __setitem__(self, key: str, value: VariableRepr):
-        """Sets a single variable representation within the dataframe.
+        self._columns = []
+
+    @property
+    def columns(self):
```
junjiejiangjjj, line 45: Implemented functions must have unit tests

junjiejiangjjj (changes requested):
Implemented functions must have unit tests

Chiiizzzy:
All problems fixed, added ut for each function @fzliu @junjiejiangjjj

towhee/dag/dataframe_repr.py
```diff
@@ -12,83 +12,154 @@
...
+        return self._columns
+
+    @columns.setter
+    def columns(self, value: List[VariableRepr]):
+        self._columns = value
 
-    def __getitem__(self, key) -> VariableRepr:
-        """Access a variable representation via dictionary-like indexing.
+    @staticmethod
+    def is_format(info: list):
```
junjiejiangjjj, line 50: return True/False

junjiejiangjjj (approved):
/approve
/lgtm

PR #72, opened 2021-09-10
 add scheduler for optimization

oneseer:
add scheduler.py and test_scheduler.py
@fzliu


oneseer:
import torch raise runtime error:
torch overrides.py '_has_torch_function' already has a docstring
check this issue

PR #74, opened 2021-09-11
check import torch bug

oneseer:
test if runtime error is triggered by 'import torch'


oneseer:
fixed conflict between coverage and pytorch

PR #76, opened 2021-09-12
Impl dataframe

junjiejiangjjj:
#58 Impl dataframe and map iterator


towhee/dataframe/dataframe.py
```diff
@@ -35,19 +37,171 @@ def __init__(self, name: str, data: List[Tuple[Variable]] = None):
...
+
+            if force_size and not self._end and self._total - start < count:
+                return False, []
+
+            real_count = min(self._total - start, count)
+            real_start_index = start - self._start_index
+            return self._end and real_start_index + real_count >= self._total, \
+                self._data[real_start_index: real_start_index + real_count]
+
+    def get_size(self, start):
```
GuoRentong, line 88: I suggest to use @property size

towhee/dataframe/dataframe.py
```diff
@@ -35,19 +37,171 @@ def __init__(self, name: str, data: List[Tuple[Variable]] = None):
...
+        assert start >= self._start_index
+        return self._total - start
+
+    def set_end(self):
```
GuoRentong, line 92: `def seal(self)` is a common name for this functionality

GuoRentong: and replace `self._end` with `self._sealed`

towhee/dataframe/dataframe.py
```diff
@@ -35,19 +37,171 @@ def __init__(self, name: str, data: List[Tuple[Variable]] = None):
...
         self._name = name
-        self._data = data
+        self._data = data if data is not None else []
+        self._registered_iter = []
+        self._start_index = 0
+        self._total = 0
+        self._end = False
+
+        self._lock = threading.Lock()
+        self._registered_lock = threading.Lock()
```
fzliu, line 48: What is the purpose of this registered lock?

towhee/dataframe/dataframe.py
```diff
@@ -35,19 +37,171 @@ def __init__(self, name: str, data: List[Tuple[Variable]] = None):
...
+        Get [start: start + count) items of dataframe
+        `force_size` is true, return data size must equal `count`. If there is
+        not enough data, return empty list.
+        `force_size` is false, return no more than `count` data
+
+        Returns:
+            (is_end, data)
+        """
+        # TODO (junjie.jiang) call gc function
+        with self._lock:
```
GuoRentong, line 71: we need further optimization for`get` `put` `seal` based on the following logic. (we refactor this part of codes after v0.1)
* put and seal will be called by the same operator
* seal will always be called after put
* iterators never ends until they meet `df._sealed == True`

towhee/dataframe/dataframe.py
GuoRentong: and provides one addtional methods `is_sealed(self)->bool`

GuoRentong (changes requested):

towhee/dataframe/dataframe.py
```diff
@@ -35,19 +37,171 @@ def __init__(self, name: str, data: List[Tuple[Variable]] = None):
...
+        # self._n = n
+        # self._i = 0
+        raise NotImplementedError
 
-    def append(self, item: Tuple[Variable]):
-        self.data.append(item)
+    def __next__(self):
+        if self._i >= self._n:
+            raise StopIteration
+        return self._data[0]
```
fzliu, line 207: There should be a check here to ensure that the dataframe is not empty.

towhee/dataframe/dataframe.py
```diff
@@ -35,19 +37,171 @@ def __init__(self, name: str, data: List[Tuple[Variable]] = None):
...
+        self._registered_lock = threading.Lock()
 
     @property
     def name(self):
         return self._name
 
-    def __getitem__(self, val):
-        return self.data[val]
+    def put(self, item: Tuple[Variable]) -> None:
+        assert not self._end, 'DataFrame %s already end, can not put data' % self._name
```
fzliu, line 55: Let's raise a `ValueError` or some other custom `DataframeError` (instead of `AssertionError`) here so that it is clearer to the user exactly what the issue is.

towhee/dataframe/dataframe.py
```diff
@@ -35,19 +37,171 @@ def __init__(self, name: str, data: List[Tuple[Variable]] = None):
...
+            n:
+                the repeat times. If `None`, the iterator will loop continuously.
+    """
 
-    def __delete__(self, val):
-        del self.data[val]
+    def __init__(self, df: DataFrame, n: int = None):
+        # self._n = n
+        # self._i = 0
+        raise NotImplementedError
```
fzliu, line 202: This iterator looks mostly fine to me, we can remove this error and deliver it with v0.1 (adding tests later).

towhee/dataframe/dataframe.py
junjiejiangjjj: get_size function need start param to calc len(dataframe[start: end]), @property function can not has params except self

towhee/dataframe/dataframe.py
junjiejiangjjj: Add TODO for this

towhee/dataframe/dataframe.py
junjiejiangjjj: Only implement map iterator, others will implement after v0.1

towhee/dataframe/dataframe.py
junjiejiangjjj: Only implement map iterator, others will implement after v0.1

towhee/dataframe/dataframe.py
junjiejiangjjj: A dataframe's iter will register in multithread, so I use `_registered_lock` to make thread-safe

towhee/dataframe/dataframe.py
junjiejiangjjj: We can define some custom errors after v0.1

towhee/dataframe/dataframe.py
GuoRentong: why not letting the caller handle the work `df.size - start`?

GuoRentong (changes requested):

PR #83, opened 2021-09-14
WIP: Refactor Array and DataFrame

GuoRentong:
* use `list` as item container in `Array`
* remove gc related functionalities in `DataFrame` and `DataFrameIterators`
* use `Condition` to handle the data append and data access between `DataFrame` and `DataFrameIterators`

### DIFF
--- a/towhee/dataframe/dataframe.py
+++ b/towhee/dataframe/dataframe.py
@@ -13,152 +13,162 @@
 # limitations under the License.
 
 import threading
-from typing import Callable, List, Tuple
+from typing import Callable, List
 import weakref
 
-from towhee.dataframe.variable import Variable
+from towhee.array import Array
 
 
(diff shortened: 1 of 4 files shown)


towhee/dataframe/dataframe.py
```diff
@@ -13,151 +13,181 @@
...
         """
-        Get [start: start + count) items of dataframe
-        `force_size` is true, return data size must equal `count`. If there is
-        not enough data, return empty list.
-        `force_size` is false, return no more than `count` data
+        assert not self._sealed, 'DataFrame %s is already sealed, can not put data' % self._name
+
+        self._wcond.acquire()
+        try:
+            for i, v in row.items():
```
junjiejiangjjj, line 72: list object has no attribute `items`

PR #85, opened 2021-09-16
add scheduler for optimizaiton and ut

oneseer:
shceduler.py handles the learning rate of algorithm optimization


oneseer:
need to add numpy

PR #87, opened 2021-09-16
op ctx impl

junjiejiangjjj:
#58 
1. op context implementation
2. Adjust test util
3. Task implementation


towhee/engine/task.py
```diff
@@ -45,14 +45,15 @@ class Task:
     """
 
     def __init__(self, op_name: str, hub_op_id: str, op_args: Dict[str, Any],
-                 inputs: Dict[str, Any], task_idx: int):
+                 inputs: Dict[str, Any], task_idx: int, output_callback: Callable = None):
```
fzliu, line 48: Instead of having an output callback, it makes more sense to have a task finish callback which reads `task.outputs`.

PR #90, opened 2021-09-17
add adafactor and adamw optimizer

oneseer:
add adafactor.py
add adamw.py
add ut for adafactor and adamw in test_scheduler.py


oneseer:
towhee/tests/operators/test_operator_base.py:41:0: C0304: Final newline missing (missing-final-newline)

PR #91, opened 2021-09-17
add adafactor and adamw optimizer

oneseer:
add adafactor.py and its ut in test_scheduler
add adamw.py and its ut in test_scheduler
fix C0304: Final newline missing (missing-final-newline) in test_opeator_base.py


oneseer:
merge the three commits

PR #92, opened 2021-09-17
add adafactor and adamw 

oneseer:
add adafactor.py and its ut
add adamw.py and its ut
fix test_operator_base.py


oneseer:
too many commits together

PR #93, opened 2021-09-17
add adamw and adafactor optimizer

oneseer:
add  adamw.py and its ut in test_scheduler.py
add  adafactor.py and its ut in test_scheduler.py


junjiejiangjjj (commented):
/approve
/lgtm

PR #101, opened 2021-09-22
Engine Impl

junjiejiangjjj:
Engine impl
Combined test of engine/scheduler/task-executor/task  

#58

### DIFF
--- a/towhee/engine/task_scheduler.py
+++ b/towhee/engine/task_scheduler.py
@@ -40,22 +40,26 @@ class _TaskScheduler(ABC):
             assign any further tasks to it.
     """
 
-    def __init__(self, task_execs: List[TaskExecutor], max_tasks: int = 10):
+    def __init__(self, pipelines: List[Pipeline], task_execs: List[TaskExecutor], max_tasks: int = 10):
         self._task_execs = task_execs
         self._max_tasks = max_tasks
-        self._pipelines = []
+        self._pipelines = pipelines
+        self._need_stop = False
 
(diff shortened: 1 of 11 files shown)


towhee/engine/task_scheduler.py
```diff
@@ -40,30 +40,34 @@ class _TaskScheduler(ABC):
...
 
-        Args:
-            pipeline: `towhee.Pipeline`
-                A single pipeline to schedule tasks for.
-        """
-        pipeline.add_task_ready_handler(self.on_task_ready)
-        pipeline.add_task_start_handler(self.on_task_start)
-        pipeline.add_task_finish_handler(self.on_task_finish)
-        self._pipelines.append(pipeline)
+    # def add_pipeline(self, pipeline: Pipeline):
```
fzliu, line 52: Adding new pipelines after the `Engine` is created seems like a fairly important feature. Do we plan on adding this back later?

junjiejiangjjj: task_scheduler's  `self._pipelines` is the ref of pipelines, we only need call engine's `add_pipeline` function.

fzliu: Correct, but we also need to add the `Pipeline`'s task handlers, which we won't be able to do without wrapping it in a function.

junjiejiangjjj: In v0.1， we only have `write_output` handler to get one task's outputs. After v0.1, let us consider how to deal with other handlers.

PR #102, opened 2021-09-23
add transformer.py for transformer modelling

oneseer:
add test_transformer.py for transformer ut


oneseer:
need to fix docstring of transformer.py

PR #128, opened 2021-09-26
add pytorch transform operator

oneseer:
add pytorch_transform_operator.py
add test_pytorch_transform_operator.py

### DIFF
--- a/towhee/tests/mock_operators/__init__.py
+++ b/towhee/tests/mock_operators/__init__.py
@@ -25,6 +25,7 @@ _MOCK_OPERATOR_DIR = os.path.dirname(__file__)
 ADD_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "add_operator")
 PYTORCH_CNN_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "pytorch_cnn_operator")
 PYTORCH_TRANSFORMER_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "pytorch_transformer_operator")
+PYTORCH_TRANSFORME_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "pytorch_transform_operator")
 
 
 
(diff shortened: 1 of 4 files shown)


towhee/tests/mock_operators/pytorch_transform_operator/pytorch_transform_operator.py
```diff
@@ -0,0 +1,35 @@
...
+
+class PytorchTransformOperator(Operator):
+    """
+    Image transform operator
+        Args:
+            tfms(transforms.Compose):
+                This is used to transform an image.
+    """
+
+    def __init__(self, tfms) -> None:
```
fzliu, line 29: Can we use a `str` here instead of `transforms.Compose`? Otherwise we will have trouble adding this argument into the YAML.

towhee/tests/operators/test_pytorch_transform_operator.py
```diff
@@ -0,0 +1,46 @@
...
+from torchvision import transforms
+from torchvision import models
+import os
+
+from towhee.tests.mock_operators import PYTORCH_TRANSFORME_OPERATOR_PATH, load_local_operator
+
+
+class PytorchTransformTest(unittest.TestCase):
+    pwd = os.getcwd()
+    print('pwd is ' + str(pwd))
```
fzliu, line 27: We can leave these `print`s here for now, but let's get rid of them later on.

oneseer: ok

towhee/tests/mock_operators/__init__.py
```diff
@@ -25,6 +25,7 @@
 ADD_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "add_operator")
 PYTORCH_CNN_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "pytorch_cnn_operator")
 PYTORCH_TRANSFORMER_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "pytorch_transformer_operator")
+PYTORCH_TRANSFORME_OPERATOR_PATH = os.path.join(_MOCK_OPERATOR_DIR, "pytorch_transform_operator")
```
fzliu, line 28: There's an extra `E` after `TRANSFORM`.

Issue #154, opened 2021-09-30
Pipeline crashed with exception when there is no image


binbinlv:
Steps/Code to reproduce:
1 setup env and install towhee:
```python3
python3 -m venv towhee-test-env
pip install -U pip
pip install towhee
```
2 into python env: phthon3:
```python3
from towhee import pipeline
p = pipeline('image-embedding')
embedding = p()
```
Expected result:
Return error

Actual results:
pipeline crashed with eexception:
```
>>> embedding = p()
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/Users/binbin/Towee/towhee-test-env/lib/python3.8/site-packages/towhee/engine/engine.py", line 85, in run
    self._task_sched.schedule_forever(self._config.sched_interval_ms)
  File "/Users/binbin/Towee/towhee-test-env/lib/python3.8/site-packages/towhee/engine/task_scheduler.py", line 78, in schedule_forever
    self.schedule_step()
  File "/Users/binbin/Towee/towhee-test-env/lib/python3.8/site-packages/towhee/engine/task_scheduler.py", line 139, in schedule_step
    tasks = op_ctx.pop_ready_tasks(n_tasks=1)
  File "/Users/binbin/Towee/towhee-test-env/lib/python3.8/site-packages/towhee/engine/operator_context.py", line 99, in pop_ready_tasks
    op_input_params = self._reader.read()
  File "/Users/binbin/Towee/towhee-test-env/lib/python3.8/site-packages/towhee/engine/_operator_io.py", line 87, in read
    return self._to_op_inputs(data[0])
  File "/Users/binbin/Towee/towhee-test-env/lib/python3.8/site-packages/towhee/engine/_operator_io.py", line 63, in _to_op_inputs
    ret[key] = cols[index].value
IndexError: tuple index out of range
```

Environment:
Mac OS

fzliu:
Leaving this issue open for now (will be fixed in v0.2).

PR #163, opened 2021-10-07
Earlier operator cache check and empty input check

filip-halt:
Checking cache and downloading from hub in operator_loader will cause multiple file downloads if more than one device in the system. Instead check and download during pipeline creation. Logic for checking is there, just need to figure out how we will be storing in hub.

Quick fix for https://github.com/towhee-io/towhee/issues/154 by checking if any data is given. Future work can check if the data matches the input data type, for now a log error msg is created rather than raising an error for not using the correct data type, causing an IndexingError later on.

### DIFF
--- a/towhee/__init__.py
+++ b/towhee/__init__.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from pathlib import Path
 from typing import Any, Tuple
 import os
@@ -52,6 +51,10 @@ class _PipelineWrapper:
     def __call__(self, *args) -> Tuple[Any]:
         """Wraps the input arguments around a `Dataframe` for Pipeline.__call__().
         """
+        # Check if no data supplied to pipeline
+        if not args:
+            raise ValueError(
+                'No data supplied to pipeline')
 
         # Create `Variable` tuple from input arguments.
         vargs = []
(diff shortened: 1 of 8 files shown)


towhee/__init__.py
```diff
@@ -52,6 +51,10 @@ def __init__(self, pipeline_: Pipeline):
     def __call__(self, *args) -> Tuple[Any]:
         """Wraps the input arguments around a `Dataframe` for Pipeline.__call__().
         """
+        # Check if no data supplied to pipeline
+        if len(args) == 0:
```
fzliu, line 55: A simpler way to write this is `if not args`.

towhee/engine/pipeline.py
```diff
@@ -61,6 +63,19 @@ def __call__(self, inputs: DataFrame) -> DataFrame:
...
+    # def fill_cache(self):  
+    #     """
+    #     Check to see if the operator directory exists locally or if needs to
+    #     downloaded from hub
+    #     """      
+    #     for key, value in self._graph_repr._operators.items():
+    #         if key not in ('_start_op', '_end_op'):
+    #             operator_path = LOCAL_OPERATOR_CACHE / (value.function)
+    #             if operator_path.is_dir() is False:
+    #                 print("missing", operator_path)
```
fzliu, line 76: This code won't pass `pylint`, we use single-quotes for Python strings.

fzliu (changes requested):

towhee/engine/pipeline.py
```diff
@@ -13,9 +13,9 @@
 # limitations under the License.
 
 from towhee.dag.graph_repr import GraphRepr
-
 from towhee.engine.graph_context import GraphContext
 from towhee.dataframe import DataFrame
+from towhee.engine import LOCAL_OPERATOR_CACHE
```
fzliu, line 18: Comment this out for now so `pylint` can pass.

fzliu (changes requested):

PR #168, opened 2021-10-11
Add pipeline: get image embedding and insert to milvus

jaelgu:

Add a pipeline: input image --> get embedding --> insert to milvus --> output milvus id


jaelgu:
> I think we can put our built-in pipelines in a more direct location instead of in the `tests` folder. Any suggestions? @junjiejiangjjj

Moved to /towhee/pipeline and rename the pipeline.

PR #172, opened 2021-10-12
Add operator for object detection with yolo

shiyu22:


filip-halt:
> @fzliu It shows the error `ModuleNotFoundError: No module named 'cv2'`, how can I install the package?

I think you can download it using pip install opencv-python

PR #173, opened 2021-10-12
Add operator: milvus_search

jaelgu:


filip-halt:
I think the hub is almost ready for uploads and such. It might be worth uploading them there instead of towhee itself.

jaelgu:
> I think the hub is almost ready for uploads and such. It might be worth uploading them there instead of towhee itself.

Done

jaelgu:
uploaded to hub: https://hub.towhee.io/jaelgu/milvus_search

PR #175, opened 2021-10-12
adding tools for interacting with hub

filip-halt:
Adding tools for downloading a repo through http api and for creating/deleting repos through api. Script is available at the bottom thanks to @Chiiizzzy . Basic testing done for now but more testing will need to be done once hub directory format is decided.

### DIFF
--- a/towhee/utils/hub_tools.py
+++ b/towhee/utils/hub_tools.py
@@ -1,0 +1,413 @@
+import requests
+import os
+import random
+import sys
+import getopt
+from typing import List
+
+from tempfile import TemporaryFile
+from requests.auth import HTTPBasicAuth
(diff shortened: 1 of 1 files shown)


towhee/utils/hub_tools.py
```diff
@@ -0,0 +1,414 @@
...
+            os.makedirs(os.path.dirname(f))
+        except OSError as e:
+            raise e
+    try:
+        r = requests.get(url, stream=True)
+        r.raise_for_status()
+    except HTTPError as e:
+        raise e
+
+    local_file = open(f, 'wb')
```
Chiiizzzy, line 309: Use `with open` to pass the pylint check

PR #179, opened 2021-10-14
Implement multi-thread downloads

Chiiizzzy:
- Format 
- support multi-thread download


filip-halt:
Any issues with two threads creating a directory at once? Both check and its not there and then both try to create it?

filip-halt:
Looks all good @fzliu

PR #183, opened 2021-10-16
General pipeline refactor

fzliu:
This PR implements the following high-level changes:
- Input `DataFrame`s to a `Pipeline` can now have multiple rows
- `Pipeline`s now wait properly for all outputs to be ready before returning
- Sealing an input `DataFrame` to an `Operator` now properly chains to its corresponding output `DataFrame`


towhee/engine/graph_context.py
```diff
@@ -56,35 +56,40 @@ def __init__(self, ctx_idx: int, graph_repr: GraphRepr):
         self._cv = threading.Condition()
 
     def __call__(self, inputs: Tuple):
-        graph_input = self.operator_contexts['_start_op'].outputs[0]
-        graph_input.put(inputs)
-        graph_input.seal()
+        self.op_ctxs['_start_op'].outputs[0].put(inputs)
```
GuoRentong, line 59: self.inputs.put(inputs)

towhee/engine/pipeline.py
```diff
@@ -26,65 +27,91 @@ class Pipeline:
...
+        for n in range(self._parallelism):
+            graph_ctx = GraphContext(n, self._graph_repr)
+            self._scheduler.register(graph_ctx)
+            graph_ctxs.append(graph_ctx)
+
+        # Weave inputs into each GraphContext.
+        # TODO(fzliu): can we spin off a thread to do this?
+        for n, row in enumerate(inputs.map_iter()):
+            idx = n % len(graph_ctxs)
+            graph_ctxs[idx](row[0])
```
GuoRentong, line 88: a graph_ctxs can not process multiple rows simultaneously.
if input rows > len(graph_ctxs), will a graph context being called again before it finish its last call?

GuoRentong (changes requested):

towhee/engine/graph_context.py
fzliu: Fixed.

towhee/engine/pipeline.py
fzliu: Made `_start_op` and `_end_op` schedulable operators and added callbacks to set a `GraphContext`'s `_is_busy` flag. When `_is_busy` evaluates to True, new inputs to `_start_op` disabled.

towhee/engine/operator_context.py
```diff
@@ -135,9 +134,8 @@ def num_finished_tasks(self) -> int:
         # self._finished_tasks' only monifier.
 
     def _write_outputs(self, task: Task):
```
junjiejiangjjj, line 136: This function will be called in task-executor，so we must use the lock to make thread-safe

towhee/engine/pipeline.py
```diff
@@ -26,65 +27,91 @@ class Pipeline:
...
+
+        # Instantiate an output `DataFrame`. Upon completion, individual `GraphContext`
+        # outputs are merged into this dataframe.
+        # TODO(fzliu): Create an operator to merge multiple `DataFrame` instances.
+        for graph_ctx in graph_ctxs:
```
junjiejiangjjj, line 93: 1. How to make the output sequence and the input sequence have the same order
2. For each input, how to reset ctx
3. We should have tests for multi inputs

PR #199, opened 2021-10-18
Hub Integration: Check Cache and Download

filip-halt:
Remade https://github.com/towhee-io/towhee/pull/180 due to editor bug.


Chiiizzzy:
looks all good. 
But I suppose we will delete pipeline.yaml and mock_operators from `/tests` once the hub is all set, so we might to change the way we load default pipeline in `engine/__init__.py` in the future.

towhee/engine/__init__.py
```diff
@@ -22,4 +22,6 @@
 MOCK_PIPELINES = str(Path(__file__).parent.parent /
                      'tests/test_util/resnet50_embedding.yaml')
 
+# These are used for loading defaults into .towhee
 MOCK_OPS = str(Path(__file__).parent.parent / 'tests/mock_operators')
+MOCK_PIPES = str(Path(__file__).parent.parent / 'tests/mock_pipelines')
```
Chiiizzzy, line 27: We might remove `mock_piplines` and `mock_operators` once hub is all set, so we might need to change this in the future.

PR #206, opened 2021-10-19
add multi scale transformer block layer

oneseer:
add multi_scale_transformer_block.py
add test_multi_scale_transformer_block.py


fzliu:
Filename should be `multi_scale_transformer_block.py` (missing the `l` in multi).

PR #296, opened 2021-11-29
Multi-read + concat

filip-halt:
Changes to include multi-column readers. 
1. Need better testing for filter.
2. Switch to iterator blocking mechanism.
3. Do blocking for batch iterator.


towhee/engine/operator_context.py
```diff
@@ -52,13 +52,19 @@ def __init__(
...
+        for x in op_repr.inputs:
+            # If a dataframe is already added, append the new columns.
+            if x['df'] in reader_inputs.keys():
+                reader_inputs[x['df']]['cols'].append((x['name'], x['col']))
+            else:
+                reader_inputs[x['df']] = {}
+                reader_inputs[x['df']]['df'] = dataframes[x['df']]
+                reader_inputs[x['df']]['cols'] = [(x['name'], x['col'])]
+
+        self._reader = create_reader(reader_inputs, iter_type)
```
filip-halt, line 67: @junjiejiangjjj How do you feel about this change. I think that moving to a dictionary that stores all the info about the input to be easier to use and adapt. Its structure is like this. 
{'_start_df': 
                    {'cols': [('num1', 0), ('num2', 1), ('num3', 2)],
                     'df': <towhee.dataframe.dataframe.DataFrame object at 0x10f48e8b0>}}

towhee/engine/operator_io/reader.py
```diff
@@ -90,18 +89,35 @@ def read(self) -> Tuple[Dict[str, any], Tuple]:
...
+                    # self._iters[_][0] is the iterator
+                    # self._iters[_][1] are the columns for that iterator
+                    data = next(self._iters[iter_index][0])
+                    if data is not None:
+                        for (key, index) in self._iters[iter_index][1]:
+                            ret[key] = data[index].value
+                        origin_data += data
+                        remove_index.add(bit_index)
+                bitmap = [i for j, i in enumerate(bitmap) if j not in remove_index]
+            return ret, origin_data
```
filip-halt, line 114: This reader blocks by looping to get the remaining values. Should it rather use the new block feature in the iterator?

towhee/tests/mock_operators/add_operator_v2/add_operator_v2.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Add operator that can take in undefined amount of inputs and saves results to '<input_name>_sum'
+    column of result dataframe.
+    """
+
+    def __init__(self, factor: int) -> None:
+        super().__init__()
+        self._factor = factor
+
+    def __call__(self, **kwargs):
```
junjiejiangjjj, line 30: If the inputs and the outputs are dynamic, how should I create a pipeline without knowing the operator codes?

junjiejiangjjj: We will use python interface which likes spark to create pipelines, so we must know the inputs and outputs col keys without the operator's codes.

junjiejiangjjj: Filter and concat may directly use previous operator's col-keys, but other operators, I think must have their own col-keys.

towhee/tests/mock_operators/concat_operator/concat_operator.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    def __init__(self, repeat_end: bool) -> None:
+        super().__init__()
+        self._scalar = repeat_end
+        self._last_seens = defaultdict(lambda: None)
+
+    def __call__(self, **kwargs):
+        ret = []
+        none_count = 0
```
fzliu, line 34: `if not (set(kwargs.values()) - {None}): pass` seems more pythonic here instead of using `none_count`. Performance overhead should be minimal.

towhee/dataframe/dataframe.py
```diff
@@ -213,6 +213,7 @@ class DataFrameIterator:
     def __init__(self, df: DataFrame):
         self._df_ref = weakref.ref(df)
         self._cur_idx = 0
+        self._df_name = df._name
```
fzliu, line 216: Directly accessing variables beginning with an underscore is bad practice.

junjiejiangjjj:
We will refactor `dataframe` soon, one row will not be a tuple, but a struct, we need to add additional info such as row id, timestamp, and other info we need.

towhee/engine/operator_context.py
junjiejiangjjj: > 

`OrderedDict` can keep the order always the same, but can not control the order. If we want df1's row always before df2's row if  their  have same row key, `List` can do this, but `OrderedDict` can't.

towhee/tests/mock_operators/concat_operator/concat_operator.py
```diff
@@ -0,0 +1,51 @@
...
+    Stateful concat operator that can perform scalar and nonscalar concat. If no new data
+    is presented, the row will be skipped.
+    """
+
+    def __init__(self, repeat_end: bool) -> None:
+        super().__init__()
+        self._scalar = repeat_end
+        self._last_seens = defaultdict(lambda: None)
+
+    def __call__(self, **kwargs):
```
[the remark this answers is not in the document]
filip-halt: Yea this is only for column based

towhee/tests/mock_operators/add_operator_v2/add_operator_v2.py
filip-halt: this was more for a setup of doing multiple adds in one operator. Add operator v1 only has num, only allowing addition of one column/df.

towhee/tests/mock_operators/add_operator_v2/add_operator_v2.py
```diff
@@ -0,0 +1,42 @@
...
+        cols = []
+        ret = []
+        for key, value in kwargs.items():
+            cols.append((key + '_sum', type(value)))
+            ret.append(value + self._factor)
+
+        outputs = NamedTuple('Outputs', cols)
```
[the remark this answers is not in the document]
filip-halt: the type? i just did type here, can probably just change it to int.

towhee/tests/mock_operators/concat_operator/concat_operator.py
filip-halt: Yea, probably a better way of writing it. ill change it

towhee/dataframe/dataframe.py
```diff
@@ -152,6 +155,7 @@ def get(self, start: int, count: int, block: bool = False) -> List[Tuple[Variabl
 
     def notify_block_readers(self):
         with self._accessible_cv:
+            self._kill_cv = True
```
filip-halt, line 158: temporary fix, if block readers are killed then that dataframe wont be able to block read again. Only way to fix this is to make quite a few changes to df because we are limited to one conditional and that conditional needs to be able to tell between getting new values and being force woken up.

towhee/dataframe/dataframe.py
```diff
@@ -278,14 +285,26 @@ def __init__(self, df: DataFrame, size: int, step: int, block=False):
         self._block = block
 
     def __next__(self):
-        data = self._df_ref().get(self._cur_idx, self._size, self._block)
-        if self._df_ref().sealed and not data:
-            raise StopIteration
+        if self._block:
```
junjiejiangjjj, line 288: Can you polish it? because the code is similar.

PR #317, opened 2021-12-15
[WIP] Dataframe refactor

filip-halt:
Dataframe refactor to use GC-able arrays instead of tuples.


towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,146 @@
...
+        df = self._df_ref()
+
+        code, row = df.get(self._offset, count = 1, iter_id = self._id)
+
+        if code == 'Index_GC':
+            raise IndexError
+
+        elif code == 'Index_OOB_Unsealed':
+            if self._block:
+                while code == 'Index_OOB_Unsealed':
```
junjiejiangjjj, line 47: We couldn't use while-loops and sleep to wait for data. If one image-pipeline has 10 ops, the worst case we may sleep 10s, but one image maybe only need 100ms.

junjiejiangjjj: And if a big pipeline has hundreds of iters, the pipeline-process will do nothing but while-loops.

filip-halt: Is there a good tactic for this other than condition vars? Condition vars wont transfer well to the distributed version. For the waiting I only set it to 1s for testing purposes. It can be much lower, just needs a balance between checking to much and waiting too long. 

Furthermore, we will probably want to keep this logic within the iterator and not within dataframe.

junjiejiangjjj: We can't use while-loops if we case about latency.

towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        if self._done:
+            raise StopIteration
+
+        df = self._df_ref()
+
+        code, row = df.get(self._offset, count = self._batch_size, iter_id = self._id)
+        print(code)
+
+        if code == 'Index_GC':
```
junjiejiangjjj, line 115: We could use the emun to define the return code.

towhee/dataframe/dataframe_v2.py
```diff
@@ -33,18 +33,41 @@ class DataFrame:
...
-        self._lock = threading.Lock()
+
+        self._iterator_lock = threading.Lock()
+        self._it_id = 0
+        self._iterators = []
+        self._iterator_offsets = []
+        self._min_offset = 0
+
+        self._block_lock = threading.Lock()
+        self._blocked = {}
```
junjiejiangjjj, line 50: Maybe we could put the `condition` into iterators, that means iterator has a `condition` and a `notify` function, if iterator found there is no data, it calls `condition` wait. In dataframe's `put`, we can call iter's `notify` to wake up the iter.

junjiejiangjjj: Maybe the `Event` is enough.

filip-halt:
```python 
elif code == 'Index_OOB_Unsealed':
            if self._block:
                cv = df.notify_block(self._offset, self._batch_size)
                with cv:
                    cv.wait()
                return self.__next__()

            return None

```

How about this? If the iterator needs to block it registers itself that it needs to block. In the df there is a dict that holds keeps track of needed indexes and a corresponding condition. 

```python 
def notify_block(self, offset, count):
        index = offset + count
        if not self._blocked.get(index):
            self._blocked[index] = threading.Condition()
        return self._blocked[index]
```

And then during a put it checks if there exists a cv waiting on that value. 

```python
self._len += 1

            cv = self._blocked.pop(self._len, None)

            if cv is not None:
                cv.notfiy_all()
```

PR #321, opened 2021-12-17
Allow specifying framework either in YAMl or in pipeline args

Chiiizzzy:
Since we want to support framework specification in both pipeline initialization and pipeline YAML, we need to pass framework arg in operators' __init__() function and modify system code accordingly. 

1. Add arg `framework` in pipelines' __init__ function, so users can specify framework when initialize a pipeline.
2. Add a member `_framework` in `Engine` as a property, since engine is a singleton, we cannot initialize it twice, system will set engine's framework every time run a pipeline accordingly.
3. Add a member `_framework` in `ThreadPoolTaskExecutor`, same as `Engine`.
3. Judge an operator before load and execution, if it's an NNOperator, add arg `framework`.
4. Add arg `framework` in `OperatorPool.acquire_operator()`
5. Add arg `framework` in `OperatorLoader.load_operator()`
6. Move `NNOperator` and `PyOperator` to `base.py`

### DIFF
--- a/towhee/__init__.py
+++ b/towhee/__init__.py
@@ -76,20 +76,20 @@ class _PipelineWrapper:
         return res
 
 
-def pipeline(task: str, fmc: FileManagerConfig = FileManagerConfig(), branch: str = 'main', force_download: bool = False):
+def pipeline(src: str, branch: str = 'main', framework: str = 'pytorch', force_download: bool = False):
     """
-    Entry method which takes either an input task or path to an operator YAML.
+    Entry method which takes either an input task name or path to an operator YAML.
 
-    A `Pipeline` object is created (based on said task) and subsequently added to the
+    A `Pipeline` object is created (based on the src) and subsequently added to the
     existing `Engine`.
 
     Args:
-        task (`str`):
+        src (`str`):
             Task name or YAML file location to use.
-        fmc (`FileManagerConfig`):
-            Optional file manager config for the local instance, defaults to local cache.
         branch (`str`):
             Which branch to use for operators/pipelines on hub, defaults to `main`.
+        framework (`str`):
+            The framework to apply.
         force_download (`bool`):
             Whether to redownload pipeline and operators.
 
@@ -99,14 +99,16 @@ def pipeline(task: str, fmc: FileManagerConfig = FileManagerConfig(), branch: st
     """
     start_engine()
 
-    if os.path.isfile(task):
-        yaml_path = task
+    if os.path.isfile(src):
+        yaml_path = src
     else:
+        fmc = FileManagerConfig()
         fm = FileManager(fmc)
-        task = DEFAULT_PIPELINES.get(task, task)
-        yaml_path = fm.get_pipeline(task, branch, force_download)
+        src = DEFAULT_PIPELINES.get(src, src)
+        yaml_path = fm.get_pipeline(src, branch, force_download)
 
     engine = Engine()
+    engine.framework = framework
     pipeline_ = Pipeline(str(yaml_path))
     engine.add_pipeline(pipeline_)
 
--- a/towhee/engine/engine.py
+++ b/towhee/engine/engine.py
@@ -26,7 +26,6 @@ class EngineConfig:
     """
     Global engine config
     """
-
     def __init__(self):
         self._sched_type = 'basic'
         self._cache_path = None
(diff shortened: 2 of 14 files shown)


fzliu:
Please add a description for this PR.

fzliu (changes requested):

towhee/engine/engine.py
```diff
@@ -65,18 +64,33 @@ class Engine(threading.Thread):
...
         self._setup_execs()
         self._setup_sched()
 
+    @property
+    def framework(self):
+        return self._framework
+
+    @framework.setter
+    def framework(self, framework: str):
+        '''
```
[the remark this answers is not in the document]
Chiiizzzy: Which part? I compared it with our coding style, seems correct.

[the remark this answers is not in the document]
fzliu: Use `"""` instead of `'''`.

towhee/__init__.py
```diff
@@ -76,7 +76,7 @@ def __call__(self, *args) -> List[Tuple]:
         return res
 
 
-def pipeline(task: str, fmc: FileManagerConfig = FileManagerConfig(), branch: str = 'main', force_download: bool = False):
+def pipeline(task: str, framework: str = 'pytorch', fmc: FileManagerConfig = FileManagerConfig(), branch: str = 'main', force_download: bool = False):
```
junjiejiangjjj, line 79: 1. `fmc` can be removed for the user does not care about it.
2. `task` in there, means pipeline's repo name, so I think we should change the name.
3.  `task` `branch` `framework` `foce_download` would be better than the current order.

towhee/__init__.py
```diff
@@ -107,6 +109,7 @@ def pipeline(task: str, fmc: FileManagerConfig = FileManagerConfig(), branch: st
         yaml_path = fm.get_pipeline(task, branch, force_download)
 
     engine = Engine()
+    engine.framework = framework
```
junjiejiangjjj, line 112: `framework` is a pipeline attribute, if we put it into `engine`, it will be wrong if I run two or more pipelines with different frameworks.

Chiiizzzy: I add a step in `towhee/__init__.py`,  every time we start a new pipeline, we will set the `framework` for `engine`.

towhee/__init__.py
Chiiizzzy: Fixed.

towhee/__init__.py
fzliu: Can we modify the contents of the `GraphRepr` passed to the `Engine` instead of modifying the `Engine` itself?

Chiiizzzy: `GraphRepr` only contains the information of the YAML, if we want to specify the framework in `pipeline()`,  we cannot avoid modifying `Engine`.

junjiejiangjjj: > I add a step in `towhee/__init__.py`, every time we start a new pipeline, we will set the `framework` for `engine`.

What if we start new pipelines in different threads at the same time?

Chiiizzzy: Shall we introduce a lock for such behavior?

junjiejiangjjj: Can we put it into `Pipeline`? Different pipelines can have their own framework info.

Chiiizzzy: Oops, that's a bad idea, or shall we change the `_pipeline` list in `Engine` to a dict to store the framework for each pipeline.

Chiiizzzy: Putting it into `Pipeline` cannot pass the framework arg to op_loader

towhee/engine/thread_pool_task_executor.py
```diff
@@ -46,6 +45,7 @@ def __init__(self, name: str, cache_path: str = None):
         self._thread_pool = ThreadPoolExecutor()
         self._is_run = True
         self.setDaemon(True)
+        self._framework = 'pytorch'
```
junjiejiangjjj, line 48: Why does the executor have framework info? It's only the pipeline or operator's attribute, not the executor's.

towhee/engine/operator_pool.py
```diff
@@ -92,7 +94,7 @@ def acquire_op(self, hub_op_id: str,
                 self._all_ops[op_key] = storage
 
             if not storage.op_available():
-                op = self._op_loader.load_operator(hub_op_id, op_args)
+                op = self._op_loader.load_operator(hub_op_id, op_args, framework)
```
junjiejiangjjj, line 97: `Operator Loader` doesn't care about the framework info, we may add framework into op_args.

PR #322, opened 2021-12-17
Add generate_repo_yaml & init_file_structure to hub_tools

shiyu22:
- When creating the repo, it will auto initialize the file structure with the template.
- Generate the YAML content of Operators according to the function's annotations.


shiyu22:
@fzliu How can I solve [it](https://github.com/towhee-io/towhee/runs/4557053026?check_suite_focus=true#step:4:33), can you [update the pylint](https://stackoverflow.com/questions/35701624/pylint-w0212-protected-access)?

towhee/hub/hub_tools.py
```diff
@@ -385,6 +388,126 @@ def download_files(user: str, repo: str, branch: str, file_list: List[str], lfs_
...
+    init_args = cls.__init__.__annotations__
+    try:
+        del init_args['return']
+    except KeyError:
+        pass
+
+    call_func = cls.__call__.__annotations__
+    try:
+        call_output = call_func.pop('return')
+        call_output = call_output._field_types
```
fzliu, line 490: Directly accessing a field beginning with an underscore will cause `pylint` to fail.

fzliu (changes requested):

fzliu:
Before merging, please rebase this into one PR.

towhee/hub/hub_tools.py
shiyu22: Actually the `_field_defaults` is [the attribute of a Namedtuple](https://pythontic.com/containers/namedtuple/_field_defaults), for example, we can get the Outputs info with it:
```python
>> from resnet_image_embedding import ResnetImageEmbedding
>>> call = ResnetImageEmbedding.__call__.__annotations__
>>> call
{'img_tensor': <class 'torch.Tensor'>, 'return': <class 'resnet_image_embedding.Outputs'>}
>>> re = call['return']
>>> re
<class 'resnet_image_embedding.Outputs'>
>>> re._field_types
{'feature_vector': <class 'numpy.ndarray'>}
```

shiyu22:
close, and opened #323.

PR #323, opened 2021-12-17
Add generate_repo_yaml & init_file_structure to hub_tools

shiyu22:
- When creating the repo, it will auto initialize the file structure with the template.
- Generate the YAML content of Operators according to the function's annotations.

### DIFF
--- a/towhee/hub/hub_tools.py
+++ b/towhee/hub/hub_tools.py
@@ -6,6 +6,8 @@ import getopt
 import time
 import subprocess
 import re
+import yaml
+from importlib import import_module
 
 from typing import List, Tuple
 from tqdm import tqdm
@@ -29,6 +31,7 @@ class Worker(Thread):
         file_name (`str`):
             The name of the file (includes extension).
     """
+
     def __init__(self, url: str, local_dir: str, file_name: str):
         super().__init__()
         self.url = url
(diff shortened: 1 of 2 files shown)


shiyu22:
In the pylint warning, actually the _field_defaults is the attribute of a Namedtuple, for example, we can get the Outputs info with it:

```bash
>>> from resnet_image_embedding import ResnetImageEmbedding
>>> call = ResnetImageEmbedding.__call__.__annotations__
>>> call
{'img_tensor': <class 'torch.Tensor'>, 'return': <class 'resnet_image_embedding.Outputs'>}
>>> re = call['return']
>>> re
<class 'resnet_image_embedding.Outputs'>
>>> re._field_types
{'feature_vector': <class 'numpy.ndarray'>}
```

So can you update the pylint?

towhee/hub/hub_tools.py
```diff
@@ -385,6 +388,126 @@ def download_files(user: str, repo: str, branch: str, file_list: List[str], lfs_
...
+    init_args = cls.__init__.__annotations__
+    try:
+        del init_args['return']
+    except KeyError:
+        pass
+
+    call_func = cls.__call__.__annotations__
+    try:
+        call_output = call_func.pop('return')
+        call_output = call_output._field_types
```
fzliu, line 490: Try `__annotations__` here.

fzliu (changes requested):

towhee/hub/hub_tools.py
shiyu22: Thanks, it works and is updated.

PR #332, opened 2021-12-22
[WIP] Dataframe Refactor

filip-halt:


filip-halt:
@junjiejiangjjj 
Moving convo to new pr from https://github.com/towhee-io/towhee/pull/317
```python 
elif code == 'Index_OOB_Unsealed':
            if self._block:
                cv = df.notify_block(self._offset, self._batch_size)
                with cv:
                    cv.wait()
                return self.__next__()

            return None

```

How about this? If the iterator needs to block it registers itself that it needs to block. In the df there is a dict that holds keeps track of needed indexes and a corresponding condition. 

```python 
def notify_block(self, offset, count):
        index = offset + count
        if not self._blocked.get(index):
            self._blocked[index] = threading.Condition()
        return self._blocked[index]
```

And then during a put it checks if there exists a cv waiting on that value. 

```python
self._len += 1

            cv = self._blocked.pop(self._len, None)

            if cv is not None:
                with cv:
                   cv.notify_all()
```

towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,152 @@
...
+        if code == 'Index_GC':
+            raise IndexError
+
+        elif code == 'Index_OOB_Unsealed':
+            if self._block:
+                cv = df.notify_block(self._id, self._offset, 1)
+                with cv:
+                    cv.wait()
+
+                return self.__next__()
```
junjiejiangjjj, line 50: Do not use recursion, if the batch size is large, it will be a problem. In some cases, users may use window-all to collect all data.

towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,152 @@
...
+                reached.
+        """
+        if self._done:
+            raise StopIteration
+
+        df = self._df_ref()
+
+        code, row = df.get(self._offset, count = 1, iter_id = self._id)
+
+        if code == 'Index_GC':
```
junjiejiangjjj, line 41: We'd better use `Enum`, not hard code strings.

towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,227 @@
...
+
+        df = self._df_ref()
+
+        code, row = df.get(self._offset, count = self._batch_size, iter_id = self._id)
+
+        if code == Responses.INDEX_GC:
+            raise IndexError
+
+        elif code == Responses.INDEX_OOB_UNSEALED:
+            if self._block:
```
junjiejiangjjj, line 121: Why not give iterator an `Event`, if the response is index_oob_unsealed, call event.wait(), and add a `notify()` function, the `dataframe` can call iter.notify() to wake up iterators.

junjiejiangjjj: It could be like this:
```python
    elif code == Responses.INDEX_OOB_UNSEALED:
        self._event.wait()
```
---
```python
  # dataframe call iterator.notify
  for iter in self._iters:
     if iter.status == 'wait':
         iter.notify()
```

filip-halt: Using an event is probably a better idea. Will still need to keep track of which index is being waited on as that one event is used both for killing the process and signifying that values are available. In dataframe v1 this was a problem because we would notify after adding one value in, and if a batch iterator was waiting, it would return None. Ill update to event design.

PR #348, opened 2021-12-26
Add encoder pipelines

fzliu:
- Add image encoding and music encoding default pipelines
- Remove arch section of README and add links to Towhee documentation


junjiejiangjjj (commented):
/approve
/lgtm

PR #369, opened 2021-12-29
Build type system(Image) and add pil_utils and ndarray_utils

Chiiizzzy:
1. Add Image class in towhee's type system
2. Add utils for pil and ndarray 
3. Add tests for the above modification


junjiejiangjjj (commented):
/approve
/lgtm

PR #426, opened 2022-01-06
Allow specifying tag of repos

Chiiizzzy:

### DIFF
--- a/towhee/hub/file_manager.py
+++ b/towhee/hub/file_manager.py
@@ -279,7 +279,7 @@ class FileManager():
                     # TODO: Figure out exception types
                     copytree(str(old_path), str(new_path))
 
-    def get_pipeline(self, pipeline: str, tag: str = 'main', install_reqs=True):
+    def get_pipeline(self, pipeline: str, tag: str, install_reqs: bool = True):
         """
         Obtain the path to the requested pipeline.
 
@@ -311,11 +311,8 @@ class FileManager():
             author = pipeline_split[0]
             repo = pipeline_split[1]
             file_name = repo.replace('-', '_')
-
             pipeline_path = self._cache_name(repo, author, tag) / (file_name + '.yaml')
-
             file_path = self._config.default_cache / 'pipelines' / pipeline_path
-            found_existing = False
 
             for path in self._config.cache_paths:
                 path = path / 'pipelines' / pipeline_path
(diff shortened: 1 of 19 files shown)


towhee/hub/file_manager.py
```diff
@@ -311,13 +311,12 @@ def get_pipeline(self, pipeline: str, tag: str = 'main', install_reqs=True):
...
-
             pipeline_path = self._cache_name(repo, author, tag) / (file_name + '.yaml')
-
             file_path = self._config.default_cache / 'pipelines' / pipeline_path
-            found_existing = False
 
-            for path in self._config.cache_paths:
+            found_existing = bool(file_path.is_file())
+
+            for path in reversed(self._config.cache_paths):
```
filip-halt, line 319: Adding a cache path adds it to the front of the list so reverse can be avoided, this change would look through in the wrong order.

towhee/hub/file_manager.py
```diff
@@ -311,13 +311,12 @@ def get_pipeline(self, pipeline: str, tag: str = 'main', install_reqs=True):
...
             repo = pipeline_split[1]
             file_name = repo.replace('-', '_')
-
             pipeline_path = self._cache_name(repo, author, tag) / (file_name + '.yaml')
-
             file_path = self._config.default_cache / 'pipelines' / pipeline_path
-            found_existing = False
 
-            for path in self._config.cache_paths:
+            found_existing = bool(file_path.is_file())
```
filip-halt, line 317: bool(file_path.is_file()) as an unnecessary extra file check.

towhee/hub/file_manager.py
```diff
@@ -345,7 +345,7 @@ def get_pipeline(self, pipeline: str, tag: str = 'main', install_reqs=True):
 
         return file_path
 
-    def get_operator(self, operator: str, tag: str = 'main', install_reqs=True):
+    def get_operator(self, operator: str, tag: str = 'main', install_reqs: bool = True):
```
filip-halt, line 348: same edits as previous

towhee/hub/file_manager.py
Chiiizzzy: I thought we are searching from the default cache path.

towhee/hub/file_manager.py
Chiiizzzy: Fixed

towhee/hub/file_manager.py
filip-halt: for path in reversed(self._config.cache_paths):
Was talking about this part, dont need the reversed. Cache path is searched in opposite order added with the default being the last one. If user adds c1, c2, c3, cache paths list would be [c3, c2, c1] with default being c1. Default is the last checked.

junjiejiangjjj:
If we need access repo, use `RepoManager`, so first, finish the `RepoManger`, and use it instead of the `git` package.

PR #443, opened 2022-01-10
DF_v2

filip-halt:


towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,222 @@
+import weakref
+
+from towhee.dataframe.dataframe_v2 import DataFrame, Responses
+
+class BaseIterator:
```
junjiejiangjjj, line 5: The code of iters is so similar, can we polish it?

towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,222 @@
...
+        df = self._df_ref()
+
+        code, row = df.get(self._offset, count = 1, iter_id = self._id)
+
+        if code == Responses.INDEX_GC:
+            raise IndexError
+
+        elif code == Responses.INDEX_OOB_UNSEALED:
+            if self._block:
+                cv = df.notify_block(self._id, self._offset, 1)
```
junjiejiangjjj, line 60: We needn't return a `Condition`, just add an `Event` in iter's `__init__`, and there, we can call sef._event.wait()

towhee/dataframe/dataframe_v2.py
```diff
@@ -56,23 +57,51 @@ def __init__(
...
+
+    def _from_none(self):
+        self._data_as_list = [Array(name=name) for name, _ in self._schema.cols]
+        self._data_as_dict = {name: self._data_as_list[i] for i, (name, _) in enumerate(self._schema.cols)}
+
+    def _set_cols(self, columns):
+        for names, types in columns:
+                self._schema.add_col(name=names, col_type = types)
+
+    def _update_schema(self):
```
junjiejiangjjj, line 96: We wouldn't change schema when putting data. The schema will not be changed after dataframe `__init__`.

PR #459, opened 2022-01-12
Add frame and schema

junjiejiangjjj:
For support `time-window` and `groupby`, we add additional info, such as row_id, timestamp, and so on.

Add _Frame and _Schema structure.
Use _Frame in op runners.


towhee/dataframe/dataframe.py
```diff
@@ -77,29 +88,35 @@ def sealed(self) -> bool:
         """
         return self._sealed
 
+    def _set_frame(self, item):
```
filip-halt, line 91: so caching and reordering will need to be handled at this step of insert. In addition to, its safe to assume that if the first value added already has a rowid, then every added value will have a row_id. Will this ever change in the future?

towhee/dataframe/_schema.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Every dataframe has _Frame cols.
+    """
+
+    def __init__(self):
+        self._cols = []
+        self._key_index = {}
+        self._sealed = False
+
+    def add_col(self, name: str, col_type: str) -> bool:
```
filip-halt, line 31: Would there be a downside to storing the type() of the col rather than a string of the type?

junjiejiangjjj: Schema's col info is set in the pipeline's YAML file, so the inputs are strings. And I think the schema shouldn't bind with a language.

PR #498, opened 2022-01-20
Add test cases

jingkl:
Signed-off-by: jingkl <jingjing.jia@zilliz.com>

### DIFF
--- a/tests/testcases/requirements.text
+++ b/tests/testcases/requirements.text
@@ -2,5 +2,6 @@
 #pip install -U pip
 #pip cache purge
 Pillow
+numpy
 towhee==0.4.1.dev31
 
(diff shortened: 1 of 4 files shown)


binbinlv:
please format the code like "nums=1000" as "nums = 1000".

binbinlv:
And please add the annotation for each test case

binbinlv:
3. And it is better to add a log in each embedding round when you embedding for many times.
like "print("embedding  for %d round" % (i+1))"

binbinlv:
4. Stress tests and performance tests which will cost a lot of time, so they are not needed to be run in CI tests.
Possible solution:

Skip the tests in "image_class_pipeline_register" and "audio_class_pipeline_register": (like)

skipped_cases = ["test_embedding_concurrent_multi_threads", "test_embedding_more_times", "test_embedding_avg_time"]

PR #517, opened 2022-01-25
Add Operator APIs to save & load weights

jaelgu:

This commit works with the test operator: https://towhee.io/towhee/train-tmp-effi-b5-op (run `python test.py`)


jaelgu:
Re-submitted by #519 since the codecov commenter does not update with results by new commits

PR #532, opened 2022-02-01
[WIP]Dataframe v2.1

filip-halt:


towhee/dataframe/dataframe_v2.py
```diff
@@ -13,207 +13,447 @@
...
+        self._data_as_list = []
+        self._schema = _Schema()
+
+        self._initialize_storage(columns)
+
+    def _initialize_storage(self, columns):
+        """Set the columns in the schema."""
+        for name, tipe in columns:
+            self._schema.add_col(name=name, col_type = tipe)
+        if self._schema.col_index('_frame') is None:
```
junjiejiangjjj, line 68: I think there is always be None, so the if is not needed.

towhee/dataframe/dataframe_v2.py
```diff
@@ -13,207 +13,447 @@
...
+        with self._data_lock:
+            # access a row
+            if isinstance(key, int):
+                return tuple(self._data_as_list[i][key] for i in range(len(self._data_as_list)))
+            # access a column
+            elif isinstance(key, str):
+                index = self._schema.col_index(key)
+                return self._data_as_list[index]
+
+    def __str__(self):
```
junjiejiangjjj, line 83: We may only print the metadata, for the main data of dataframe are images.

towhee/dataframe/dataframe_v2.py
```diff
@@ -13,207 +13,447 @@
...
+
+    def _initialize_storage(self, columns):
+        """Set the columns in the schema."""
+        for name, tipe in columns:
+            self._schema.add_col(name=name, col_type = tipe)
+        if self._schema.col_index('_frame') is None:
+            self._schema.add_col('_frame', '_Frame')
+        self._data_as_list = [Array(name=name) for name, _ in self._schema.cols]
 
     def __getitem__(self, key):
```
junjiejiangjjj, line 72: We wouldn't access data by __getitem__

towhee/dataframe/dataframe_v2.py
```diff
@@ -13,207 +13,447 @@
...
 
-    def seal(self):
-        with self._lock:
-            self._sealed = True
+        # When the iterator is forcefully unblocked using unblock_iters().
+        with self._iterator_lock:
+            if iter_id is not None and self._iterators[iter_id] == float('inf'):
+                return Responses.KILLED, None
+
+        # If there is no schema that means that initialization broke at some point.
```
junjiejiangjjj, line 172: If init failed, we may raise an exception, and stop running

towhee/dataframe/iterators.py
```diff
@@ -0,0 +1,199 @@
...
+        df (`towhee.Dataframe`):
+            The dataframe that is being iterated.
+        window_size (`int`):
+            How large of a window.
+        use_timestamp (`bool`):
+            Whether to use timestamp instead of row_id.
+        block (`bool`):
+            Whether to block when data not present.
+    """
+    def __init__(self, df: DataFrame, window_size = 1, use_timestamp = False, block = True):
```
junjiejiangjjj, line 137: TimeWindowIterator's init maybe like this
```python
     def __init__(self, df: DataFrame, time_range_sec: int, time_step_sec: int, block=False):
        '''
        time_range_sec: window range
        time_step_sec: window step
        '''
        pass
```

towhee/dataframe/dataframe_v2.py
filip-halt: What about the situation where we are passing in data with _frame data already and just updating that row_id. Or should we not even include it in the schema and just use data[-1] when accessing frames.

towhee/dataframe/dataframe_v2.py
filip-halt: yes, I only included this one now for debugging purposes.

towhee/dataframe/iterators.py
filip-halt: sounds good. There might be some complexity added but it shouldnt be too bad.

Issue #542, opened 2022-02-04
[Bug]: Using Manager to create repo's creates uneditable repos.


filip-halt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When you create a repo using PipelineManager and OperatorManager, you are unable to authenticate push requests as the repo is tied to the temporary access token that was used to create it. The resulting error is a 403 when pushing.

### Expected Behavior

Create a repo with manager, push to that repo with hub.towhee.io credentials

### Steps To Reproduce

```markdown
1. Create a repo with PipelineManager
2. Clone the repo
3. Change files
4. Add, Commit, Push
5. Receive 403 error.
```

### Environment

```markdown
- Towhee version(e.g. v0.1.3 or 8b23a93):
- OS(Ubuntu or CentOS):
- CPU/Memory:
- GPU:
- Others:
```

filip-halt:
@Chiiizzzy

PR #543, opened 2022-02-04
delete repo command and basicauth for creation

filip-halt:

### DIFF
--- a/towhee/utils/hub_utils.py
+++ b/towhee/utils/hub_utils.py
@@ -197,6 +197,27 @@ class HubUtils:
         except HTTPError as e:
             raise e
 
+
+
+    def basic_create(self, password, repo_class):
+        data = {
+            'auto_init': True,
+            'default_branch': 'main',
+            'description': repo_class,
+            'name': self._repo,
+            'private': False,
+            'template': False,
+            'trust_model': 'default',
+            'type': repo_class
+        }
+        url = self._root + '/api/v1/user/repos'
+        try:
+            r = requests.post(url, data=data, auth=HTTPBasicAuth(self._author, password))
+            r.raise_for_status()
+        except HTTPError as e:
+            raise e
+
+
     def create(self, password: str, repo_class: int) -> None:
         """
         Create a repo under current account.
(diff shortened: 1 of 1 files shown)


towhee/utils/hub_utils.py
```diff
@@ -234,6 +255,29 @@ def create(self, password: str, repo_class: int) -> None:
...
+            r = requests.delete(url, headers={'Authorization': 'token ' + str(token['sha1'])})
+            r.raise_for_status()
+        except HTTPError as e:
+            if e.response.status_code == 404:
+                print('Repo not found.')
+            else:
+                raise e
+        finally:
+            self.delete_token(str(token['id']), password)
+
```
Chiiizzzy, line 280: Junjie said we don't want to expose delete to users

towhee/utils/hub_utils.py
```diff
@@ -197,6 +197,27 @@ def delete_token(self, token_id: int, password: str) -> None:
...
+            'trust_model': 'default',
+            'type': repo_class
+        }
+        url = self._root + '/api/v1/user/repos'
+        try:
+            r = requests.post(url, data=data, auth=HTTPBasicAuth(self._author, password))
+            r.raise_for_status()
+        except HTTPError as e:
+            raise e
+
```
Chiiizzzy, line 219: Why do we need this?

shiyu22: What is the difference between `create` and `basic_create` ? `basic_create` does not require a token?

filip-halt: basic create doesnt use tokens, rather the login info. I believe there is an authorization error when you try to push to a repo that was made with tokens.

towhee/utils/hub_utils.py
filip-halt: Is there a specific reason for this @junjiejiangjjj

towhee/utils/hub_utils.py
shiyu22: Is that for #542

Issue #515, opened 2022-01-25
[DesignProposal] An interface for image embedding pipelines


GuoRentong:
# Background and Motivation
currently, a pipeline developer can only create new pipelines based on yaml descriptions, which is difficult even for an experienced engineer.
To solve this problem, we can provide a high level, and more friendly interface for each typical pipeline family, for example, image embedding pipelines.
Such interface will hide most of the inner complexity, such as organizing the operator topology and writting the yaml correctly.

# Design
##  User Interface
```python
def image_embedding_pipeline(model: str|List[str] = None, ensemble: str = None, name: str = None, version: str = None):
```

params:
model: an user can pass any operators with the type of  image embedding, or a list of models;
ensemble: ensemble methods for different models;
name: pipeline name;
version: pipeline version;

## detailed design proposal
The function `image_embedding_pipeline` will do the following things via calling `ImageEmbeddingPipeline`'s methods:
* load the default yaml template of the image embedding pipelines
* replace the image embedding op based on arg `model` and `model_args`
* call `pipeline` with the modified yaml
* wrap the `Pipeline` instance into a `ImageEmbeddingPipeline` instance
* return the `ImageEmbeddingPipeline` instance

The basic methods of `ImageEmbeddingPipeline`:
```python
class ImageEmbeddingPipeline:
    def __init__(self, model, model_args):
        //todo: construct the yaml and call `pipeline` to create a raw `Pipeline`
       self.raw_pipeline_ = pipeline(...)

    def save(self, name: str, version: str, path: str):
         pass

     def push_to_hub(self):
         pass
```

reiase:
```
def image_embedding_pipeline(model='resnet101-image-embedding', model_args={args0:value0, args1:value1})
```
1. the API does not know which operator to replace;
2. `model_args` is hard to understand, we should hide meta-programming tricks from the users;

GuoRentong:
> ```
> def image_embedding_pipeline(model='resnet101-image-embedding', model_args={args0:value0, args1:value1})
> ```
> 
> 1. the API does not know which operator to replace;
> 2. `model_args` is hard to understand, we should hide meta-programming tricks from the users;

in an image embedding pipeline, their will be only one operator for image embedding, so the user do not have to select wich operator to replace, or, the user of this interface do not need to know the concept of `Operator`

PR #545, opened 2022-02-06
Add interface of image embedding pipeline

GuoRentong:
* Add `PipelineBase` as the interface of all task-oriented pipelines.
* Add interface of `ImageEmbeddingPipeline`.
* In addition, the pseudo-code of image_embedding_pipeline is given at the end of `image_embedding_pipeline.py`


GuoRentong:
#515

PR #583, opened 2022-02-10
Allow operator to resume training

jaelgu:

Example use:
- op.resume_train() --> resume training from latest checkpoint directory under `output_dir` defined in `training_config`
- op.resume_train(my_dir) --> resume training from the checkpoint directory `my_dir` instead


zc277584121 (commented):
/lgtm

PR #591, opened 2022-02-11
add image embedding pipeline

reiase:


oneseer:
the max num of models is three?

towhee/pipelines/image_embedding_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            models = model
+        num_branch = len(models)
+
+        models = dict(
+            zip([
+                'embedding_model_1',
+                'embedding_model_2',
+                'embedding_model_3',
+            ], models))
```
wxywb, line 63: it is implied that 3 embeddings ensemble is supported, if user want 4 or more, what does he need?

wxywb:
Is normalization + aggregation + meta learner will use the same string 'ensemble' to configure?

reiase:
> 

will be supported by operator ir

reiase:
support short model name in image_embedding_api

PR #640, opened 2022-02-18
PoC of DataCollection API

reiase:


towhee/engine/operator_runner/runner_base.py
```diff
@@ -170,7 +170,9 @@ def __str__(self) -> str:
 
     def _call_op(self, inputs) -> Status:
         try:
-            outputs = self._op(**inputs)
+            with param_scope() as hp:
```
junjiejiangjjj, line 173: In the engine, do not use dynamic design

PR #643, opened 2022-02-18
Replace PyYaml with yaml utils

Chiiizzzy:


reiase:
we should also drop pyyaml from requirements.txt.

Issue #703, opened 2022-02-25
[Bug]: FileManagerConfig in UT


junjiejiangjjj:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`FileManagerConfig` is a singleton, in UT, some tests add cache path, some remove, so there is a probability that UT runs failed because the cache path has been removed.

### Environment

```markdown
- Towhee version(e.g. v0.1.3 or 8b23a93):
- OS(Ubuntu or CentOS):
- CPU/Memory:
- GPU:
- Others:
```

filip-halt:
I think the only way around this is to have the tests use the same initialization of cache paths (use same setups, avoid deleting paths). The config is a singleton due to the cache paths being tied to the engine which is also a singleton.

filip-halt:
Open to any ideas on getting around such issues. Is standardizing the setup of test cases a valid solution or should we be making a bigger change to the actual framework? @junjiejiangjjj

junjiejiangjjj:
![image](https://user-images.githubusercontent.com/14136703/156523719-542581b5-7ae5-4d32-8714-bc5f58302b40.png)

Currently, we set `FileManagerConfig` in every engine test, I think we could move the code into `towhee/tests/unittests/__init__.py`   @filip-halt

filip-halt:
> ![image](https://user-images.githubusercontent.com/14136703/156523719-542581b5-7ae5-4d32-8714-bc5f58302b40.png)
> 
> Currently, we set `FileManagerConfig` in every engine test, I think we could move the code into `towhee/tests/unittests/__init__.py` @filip-halt

good idea, ill give that a shot

filip-halt:
https://github.com/towhee-io/towhee/pull/760

PR #760, opened 2022-03-03
Moving FileManager logic to init

filip-halt:


filip-halt:
Fixes #703

PR #769, opened 2022-03-04
1.1 DOC added a dublicate music recognitioin tutorial

krishnakatyal:
Signed-off-by: krishna katyal <krishna.katyal@zilliz.com>


fzliu:
Note: for testing purposes only.

PR #781, opened 2022-03-07
df ordering

filip-halt:
Inputs to df are ordered when the prev_id is not -1. Only map and filter track prev_id. 


filip-halt:
@junjiejiangjjj leaving out blocking and restricted cache size. These changes would require overaching changes in runners as if the cache is full then a blocked runner will never have a chance to put the correct value in. Seems like we first need to figure out value re-running and then implement limited cache size. This shouldnt be too hard to do as the df uses acks so we can do some setting to avoid acking on missed run.

Issue #863, opened 2022-03-18
[Enhancement]: Add data type AudioFrame


GuoRentong:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

Add data type `AudioFrame` to towhee.
It's basically an extension of numpy.ndarray.

### Why is this needed?

the reason is quite similar to #802

PR #864, opened 2022-03-18
Add type AudioFrame

GuoRentong:

#863


reiase:
the new `Image` class has been put into `towhee._types`

PR #883, opened 2022-03-22
Add to_pil from_pil

GuoRentong:
#803

### DIFF
--- a/towhee/types/image_utils.py
+++ b/towhee/types/image_utils.py
@@ -32,3 +32,38 @@ def to_image_color(img: Image, target_mode: str):
         raise ValueError('Can not convert image from %s to %s.' % (img.mode, target_mode))
 
     return Image(cv2.cvtColor(img, flag), target_mode.upper())
+
+
+def from_pil(pil_img):
+    '''
+    Convert a PIL.Image.Image into towhee.types.Image.
+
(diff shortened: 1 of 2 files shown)


towhee/types/image_utils.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from towhee._types import Image
+import numpy as np
```
reiase, line 16: numpy is not an official dependency.

towhee/types/image_utils.py
```diff
@@ -32,3 +33,35 @@ def to_image_color(img: Image, target_mode: str):
...
+
+    Returns:
+        (`towhee.types.Image`)
+            The image wrapepd as towhee Image.
+    '''
+    return Image(np.array(pil_img), pil_img.mode)
+
+
+def to_pil(img: Image):
+    """
```
reiase, line 54: this function should also support ndarray.

reiase (changes requested):

towhee/types/image_utils.py
filip-halt: should we check if image can be converted or should we allow crash at runtime?

reiase (approved):

PR #918, opened 2022-03-25
add dc.glob

reiase:


jaelgu (commented):
dc.glob('xxx') seems not work when it doesn't represent for file path/directory. What about when it represents url or a sentence in text?

PR #928, opened 2022-03-28
fix operator always reload problem

reiase:


wxywb:
@junjiejiangjjj currently the restriction of operator file naming convention could be removed because of newly added factory method, will it be a problem to exist pipeline design?

PR #933, opened 2022-03-28
ray backend + bug and issue fixes

filip-halt:

### DIFF
--- a/towhee/functional/mixins/parallel.py
+++ b/towhee/functional/mixins/parallel.py
@@ -14,13 +14,14 @@
 import concurrent.futures
 from queue import Queue
 import asyncio
+import threading
 
 from towhee.utils.log import engine_log
-from towhee.functional.option import Option, Empty
+from towhee.functional.option import Option, Empty, _Reason
 from towhee.hparam.hyperparameter import param_scope
 
 
-def map_task(x, unary_op):
+def _map_task(x, unary_op):
     def map_wrapper():
         try:
             if isinstance(x, Option):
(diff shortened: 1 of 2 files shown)


towhee/functional/mixins/parallel.py
```diff
@@ -128,59 +161,142 @@ def pmap(self, unary_op, num_worker=None, executor=None):
...
+                if isinstance(x, _PoisonPill):
+                    flag = False
+                else:
+                    yield x
 
         async def worker():
             buff = []
             for x in self:
-                if len(buff) == num_worker:
+                if len(buff) == 1:
```
reiase, line 204: should be `if len(buff) == num_worker:`?

towhee/functional/mixins/parallel.py
```diff
@@ -128,59 +161,142 @@ def pmap(self, unary_op, num_worker=None, executor=None):
...
-        >>> len(stage_2_thread_set)
-        4
-        """
         def inner():
-            nonlocal flag
-            while flag or not queue.empty():
-                yield queue.get()
-            # executor.shutdown()
+            flag = True
+            while flag:
```
reiase, line 194: I think we can drop the flag and use while true here.

towhee/functional/mixins/parallel.py
```diff
@@ -189,54 +305,143 @@ def mmap(self, *arg):
...
+        def worker_wrapper():
+            loop.run_until_complete(worker())
+
+        t = threading.Thread(target=worker_wrapper)
+        t.start()
 
         retval = [inner(queue) for queue in queues]
         return [self.factory(x) for x in retval]
 
+class _PoisonPill():
```
reiase, line 439: The name `PoisonPill` makes people confused, EOS (end of stream, or EO-X, end of something) might be a better name.

reiase (changes requested):

reiase (approved):

PR #957, opened 2022-03-30
Add split and mv datasourcemixnin to datasetmixin

shiyu22:


towhee/functional/mixins/dataset.py
```diff
@@ -37,15 +106,63 @@ def video_decode(self):
...
+        # core API already exists
         pass
 
-    def split_train_test(self):
+    def filter_data(self):
+        # core API already exists
         pass
 
+    def split_train_test(self, size: list = [0.9, 0.1], **kws):
+        """
```
reiase, line 145: we should support multiple split: train vs test, train vs validate vs test

Issue #950, opened 2022-03-30
[DesignProposal]: User API like pandas


shiyu22:
### Background and Motivation

Is it necessary to integrate the commonly used and necessary Pandas interface in DataCollection? For example:

- [pd.rename](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rename.html)
- [pd.value_counts](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.value_counts.html)
- [pd.plot](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.html)

### Design
```python
dc.rename(columns={'CODE_GENDER':'Gender','FLAG_OWN_CAR':'Car'},inplace=True)

dc.feature_summarize['column'](target='target')
dc.plot['column']()
```

shiyu22:
close #979 #958

PR #964, opened 2022-03-30
Support value_counts and plot in DC

shiyu22:


shiyu22:
#950

towhee/functional/mixins/display.py
```diff
@@ -11,21 +11,75 @@
...
+    return wrapper
+
+
+def _plot_callback(self):
+    def wrapper(_: str, index: Tuple[str], *arg, **kws):
+        # pylint: disable=import-outside-toplevel
+        # pylint: disable=unused-argument
+        import pandas as pd
+        columns = [[] for _ in index]
+        for x in self:
```
filip-halt, line 45: wont both of these callbacks consume the data of a streamed iterator? Will value_counts and plot only be supported as the last function of a DC chain like to_list()

shiyu22: Yes, it will. Maybe I will add a check(transform to upstream) for it.

shiyu22:
Another PR #968

towhee/functional/mixins/display.py
shiyu22: We call this func with `param_scope().callholder`, so it will run in the workflow as usual functions.

PR #966, opened 2022-03-31
fix stateful operator & fix from_df

reiase:
Signed-off-by: reiase <reiase@gmail.com>

### DIFF
--- a/tests/unittests/functional/test_entity.py
+++ b/tests/unittests/functional/test_entity.py
@@ -22,7 +22,7 @@ class TestEntity(unittest.TestCase):
     def test_init(self):
         d = {'a': 'A', 'b': 'B', 'c': 'C'}
         e1 = Entity()
-        e2 = Entity(d)
+        e2 = Entity(**d)
         e3 = Entity(a='A', b='B', c='C')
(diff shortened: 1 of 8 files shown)


tests/unittests/functional/test_entity.py
```diff
@@ -22,7 +22,7 @@ class TestEntity(unittest.TestCase):
     def test_init(self):
         d = {'a': 'A', 'b': 'B', 'c': 'C'}
         e1 = Entity()
-        e2 = Entity(d)
+        e2 = Entity(**d)
```
Chiiizzzy, line 25: Entity class support loading data from a dict directly.

reiase: I'm facing a problem: the argument name `target` may conflict with the keywords argument when converting a pandas data frame to entities.

towhee/operator/stateful_operator.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, name):
         super().__init__()
```
Chiiizzzy, line 52: We might need to make StatefulOperator a subclass of Operator

PR #968, opened 2022-03-31
Support feature summarize and plot in DC

shiyu22:

### DIFF
--- a/towhee/functional/mixins/display.py
+++ b/towhee/functional/mixins/display.py
@@ -12,13 +12,122 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import numpy
+import pandas as pd
+from typing import Tuple
+
 from towhee._types import Image
+from towhee.hparam import param_scope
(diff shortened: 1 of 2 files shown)


shiyu22:
#950

towhee/functional/mixins/display.py
```diff
@@ -12,13 +12,120 @@
...
+def get_data_on_columns(self, index: Tuple[str]):
+    columns = [[] for _ in index]
+    for x in self:
+        for i in range(len(index)):
+            columns[i].append(getattr(x, index[i]))
+
+    data = {}
+    for i in range(len(index)):
+        data[index[i]] = columns[i]
+    return data
```
Chiiizzzy, line 33: We could use map() and an inner function here I think

shiyu22: Maybe it can be refactored with `to_df` later, I think `to_df` will support passing parameters `columns`, right?

shiyu22: Done

PR #977, opened 2022-04-01
Add timestamp to audio_frame

junjiejiangjjj:

### DIFF
--- a/towhee/types/audio_frame.py
+++ b/towhee/types/audio_frame.py
@@ -24,11 +24,12 @@ class AudioFrame(np.ndarray):
             The audio sample rate in Hz.
     """
 
-    def __new__(cls, data: np.ndarray, sample_rate: int = None):
+    def __new__(cls, data: np.ndarray, sample_rate: int = None, timestamp: int = None):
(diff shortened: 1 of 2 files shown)


towhee/types/audio_frame.py
```diff
@@ -24,11 +24,12 @@ class AudioFrame(np.ndarray):
             The audio sample rate in Hz.
     """
 
-    def __new__(cls, data: np.ndarray, sample_rate: int = None):
+    def __new__(cls, data: np.ndarray, sample_rate: int = None, timestamp: int = None):
         # Cast `np.ndarray` to be `AudioFrame`.
         # See https://numpy.org/doc/stable/user/basics.subclassing.html for details.
         obj = np.asarray(data).view(cls)
         obj._sample_rate = sample_rate
+        obj._timestamp = timestamp
```
reiase, line 32: can we just declare `timestamp` as a public attribute?

junjiejiangjjj: `timestamp` shouldn't be changed after init, so use a `@property` function to get `timestamp`

PR #998, opened 2022-04-02
Update select for entity

shiyu22:


towhee/functional/mixins/entity_mixin.py
```diff
@@ -199,6 +199,27 @@ def inner(x):
...
+        >>> from towhee import Entity
+        >>> from towhee import DataCollection
+        >>> (
+        ...     DataCollection([Entity(a=i, b=i, c=i) for i in range(5)])
+        ...         .select('a', 'b')
+        ...         .as_raw()
+        ...         .to_list()
+        ... )
+        [(0, 0), (1, 1), (2, 2), (3, 3), (4, 4)]
+        """
```
reiase, line 217: There also needs a code example to explain the behavior without `as_raw()`

reiase (changes requested):

towhee/functional/mixins/entity_mixin.py
shiyu22: good idea

reiase (approved):

PR #1024, opened 2022-04-06
mmap -> multiple pmap

filip-halt:

### DIFF
--- a/towhee/functional/mixins/ray.py
+++ b/towhee/functional/mixins/ray.py
@@ -1,0 +1,84 @@
+from queue import Queue
+import asyncio
+import threading
+
+from towhee.utils.log import engine_log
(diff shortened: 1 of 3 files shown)


towhee/functional/mixins/ray.py
```diff
@@ -0,0 +1,84 @@
+from queue import Queue
+import asyncio
+import threading
+
+from towhee.utils.log import engine_log
+from towhee.functional.option import Option, Empty, _Reason
+from towhee.functional.mixins.parallel import EOS
+
+
+def _map_task_ray(unary_op):
```
reiase, line 10: I suggest we have a common task class that represents such async computations. The task class also can support chunked execution which is able to reduce schedule overhead.

reiase (approved):

PR #1046, opened 2022-04-08
Support show AudioFrame

GuoRentong:

* Support AudioFrame in DataCollection.show()
* Add new tablefmt 'grid'


reiase:
I think we'd better use jinja2 for html templating. The code would be much cleaner.

PR #1047, opened 2022-04-08
Add image animation tutorial

shiyu22:
Quick through the sample notebook: https://github.com/shiyu22/towhee/blob/tutorials/tutorials/anime_style_transformer.ipynb


reiase:
I think it looks better if we can grab image from camera  with `towhee.read_camera`

PR #1065, opened 2022-04-11
Added AnimeGan functionality

AniTho:
Signed-off-by: Aniket Thomas [aniketthomas27@gmail.com]


reiase:
Hi @AniTho, thank you for your contribution! 

The code looks nice to me, but I think we'd better make the code easier for the other user to follow:

1. `main.py` and `test_main.py` should be merged, and use different command-line arguments for server and client;
2. There should be command-line arguments for the IP address and the listening port; the user may run into trouble if the port is already in use or he wants remote access to the API;

By the way, we have moved the examples to another separate git repo, preparing for accepting demos and code examples that contain large binary files. You can take a look at the main branch, and try to commit the example code to [towhee-io/examples](https://github.com/towhee-io/examples). A PR in the main repo is also needed to update the reference to the examples.

If you have any questions, I'm also on the #development slack channel.

shiyu22:
Same as [PR in example](https://github.com/towhee-io/examples/pull/1), close.

PR #1072, opened 2022-04-12
Add from_ops parameter in pipeline

shiyu22:

### DIFF
--- a/towhee/engine/factory.py
+++ b/towhee/engine/factory.py
@@ -201,7 +201,7 @@ class _PipelineWrapper:
         return self._pipeline
 
 
-def pipeline(pipeline_src: str, tag: str = 'main', install_reqs: bool = True):
+def pipeline(pipeline_src: str, tag: str = 'main', install_reqs: bool = True, **kwargs):
     """
     Entry method which takes either an input task or path to an operator YAML.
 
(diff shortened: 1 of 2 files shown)


towhee/engine/factory.py
```diff
@@ -201,7 +201,7 @@ def pipeline(self) -> Pipeline:
         return self._pipeline
 
 
-def pipeline(pipeline_src: str, tag: str = 'main', install_reqs: bool = True):
+def pipeline(pipeline_src: str, tag: str = 'main', install_reqs: bool = True, from_ops: bool = False):
```
reiase, line 204: We'd better not expose inner details to the users. It will make our users confused about the API document. There are three ways to pass the `from_ops` without exposing it to the users:
1. split the `pipeline` API into interface function (without `from_ops`) and implementation function (with `from_ops`). We should call the implementation function in our framework;
2. use kwargs in the `pipeline` API, and pass `from_ops` as a keyword argument;
3. use `param_scope()`. Set the parameter in `towhee.ops` and receive it within `pipeline`.

reiase (changes requested):

towhee/engine/factory.py
shiyu22: Good idea, I think *way2* is the best way to update the API and updated it.

PR #1084, opened 2022-04-16
support pandas backend

reiase:

support to create data collection from pandas, and use `DataFrame.apply` for computation, which is suppose to be faster for numerical columns.

For example, we first create a DataFrame with pandas:

```python
import pandas as pd
df = pd.DataFrame(dict(a=range(3),b=range(3)))
```

The result DataFrame:

```
   a  b
0  0  0
1  1  1
2  2  2
```

We are able to apply operators to the DataFrame with `DataCollection`

```python
dc = towhee.from_df(df) \
    .runas_op[('a', 'b'), 'c'](func=lambda x, y: x + y)
```

```
   a  b  c
0  0  0  0
1  1  1  2
2  2  2  4
```

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -7,7 +7,7 @@ tqdm>=4.59.0
 pillow>=8.3.1
 scipy>=1.5.3
 opencv-python>=4.5.3.56
-torchmetrics>=0.7.0
+torchmetrics==0.7.0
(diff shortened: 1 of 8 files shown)


Chiiizzzy:
Shall we present the ability of manipulating dataframe directly in docstring?

PR #1089, opened 2022-04-18
Add MilvusMixin for to_milvus

shiyu22:


towhee/functional/mixins/milvus_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+    return wrapper
+
+
+class MilvusMixin:
+    """
+    Mixin about Milvus, such as load data to Milvus collection.
+    """
+    def __init__(self):
+        super().__init__()
+        self.to_milvus = param_scope().callholder(_to_milvus_callback(self))
```
reiase, line 82: will sphinx generate right document for `dc.to_milvus`? can you make a try with `python setup.py build_sphinx`

towhee/functional/mixins/milvus_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if arg is not None and len(arg) == 1:
+            collection, = arg
+        elif arg is not None and len(arg) == 2:
+            collection, batch_size = arg
+
+        if 'collection' in kws:
+            collection = kws['collection']
+        if 'batch' in kws:
+            batch_size = kws['batch']
```
reiase, line 69: we'd better force the batch_size to be a int value

towhee/functional/mixins/milvus_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from towhee.hparam import param_scope
+from ..entity import Entity
+
+
+def insert(self, index, collection, batch_size):
```
reiase, line 19: it's better to tell what the function does by its name, `_milvus_insert` should be better.

towhee/functional/mixins/milvus_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+    collection.load()
+    return milvus_mr
+
+
+def _to_milvus_callback(self):
+    # pylint: disable=consider-using-get
+    def wrapper(_: str, index, *arg, **kws):
+        batch_size = 1
+        if isinstance(index, str):
+            index = (index,)
```
reiase, line 59: The index might be empty or None, we should use all the fields in the entity by default

towhee/functional/mixins/milvus_mixin.py
reiase: and the first argument of this function should be iterable, `self` should only be used for instance methods/

towhee/functional/mixins/milvus_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+
+    for entity in self:
+        for i in range(len(index)):
+            data[i].append(getattr(entity, index[i]))
+        if len(data[0]) == batch_size:
+            mr = collection.insert(data)
+            data = [[] for _ in range(len(index))]
+            primary_keys += mr.primary_keys
+            insert_count += mr.insert_count
+            print(f'Successfully inserted {mr.insert_count} row data.')
```
reiase, line 36: Log messages should be printed into stderr. And the message should be controlled by some options.

Please consider using engine log for the message

towhee/functional/mixins/milvus_mixin.py
```diff
@@ -0,0 +1,87 @@
...
+        insert_count += mr.insert_count
+        print(f'Successfully inserted {mr.insert_count} row data.')
+
+    e = Entity(insert_count=insert_count,
+               primary_keys=primary_keys,
+               delete_count=mr.delete_count,
+               upsert_count=mr.upsert_count,
+               timestamp=mr.timestamp)
+    milvus_mr = MutationResult(e)
+    collection.load()
```
reiase, line 50: what if something goes wrong, and we execute `collection.insert` without `collection.load`

reiase (changes requested):

towhee/functional/mixins/milvus_mixin.py
shiyu22: No, I haven't added `MilvusMixin` to *mixins.rst* and will add it later.

And another question, there is no example for `to_milvus`, how about installing milvus in CI and add doctest for it.

towhee/functional/mixins/milvus_mixin.py
shiyu22: good iead

towhee/functional/mixins/milvus_mixin.py
shiyu22: Do you mean the entity in DC? Maybe we need `dc.keys()` to get all field names.

towhee/functional/mixins/milvus_mixin.py
shiyu22: So finally call `collection.load` with `try-exception`.

towhee/functional/mixins/milvus_mixin.py
reiase: > Do you mean the entity in DC?

yes, we should use the schema of the entities in DC

towhee/functional/mixins/milvus_mixin.py
reiase: > No, I haven't added `MilvusMixin` to _mixins.rst_ and will add it later.
> 
> And another question, there is no example for `to_milvus`, how about installing milvus in CI and add doctest for it.

I think we should avoid heavy operations in the CI workflow.

towhee/functional/mixins/milvus_mixin.py
shiyu22: Updated with `engine_log.info`, but it will not print without if not set `engine_log.setLevel('INFO')`. Another PR will submit to add `towhee.set_log_levle` for configuration.

reiase:
/lgtm
/approve

we will check the unit tests after some cleanup @shiyu22  @junjiejiangjjj

PR #1092, opened 2022-04-18
add a serve api for data collection

reiase:

```python
import towhee
import uvicorn
from fastapi import FastAPI

app = FastAPI()

with towhee.DataCollection.api('/app1') as api:
    api.map(lambda x: x+' -> 1') \
            .map(lambda x: x+' => 1') \
            .bind(app)
with towhee.DataCollection.api('/app2') as api:
    api.map(lambda x: x+' -> 2') \
            .map(lambda x: x+' => 2') \
            .bind(app)

uvicorn.run(app)
```

test the server
```shell
curl -d 'x' http://127.0.0.1:8000/app1
```


junjiejiangjjj:
What if the `x` is ndarray, how can I use it.

PR #1136, opened 2022-04-24
Add defalut search parameter in milvus-search

shiyu22:


towhee/hub/builtin/operators/milvus_search.py
```diff
@@ -67,6 +72,20 @@ def __call__(self, query: list):
                 if schema.dtype in (101, 100):
                     self.kwargs['anns_field'] = schema.name
 
+        if 'limit' not in self.kwargs:
+            self.kwargs['limit'] = 10
+
+        index_type = self.collection.indexes[0].params['index_type']
+        if 'param' not in self.kwargs:
```
reiase, line 79: It's better to use a dict for the default parameters

reiase (changes requested):

towhee/hub/builtin/operators/milvus_search.py
shiyu22: good idea and updated it.

PR #1138, opened 2022-04-24
add tsm model and unitest

gexy185:
CBebhTuQMe
4j3yxrPDVy


jaelgu (commented):
@gexy185 Some suggested changes:
- Replace printout messages with log if possible
- Split & refactor some layers/modules by using towhee.models.layers or towhee.models.utils
- Should Mobilenet be a separate model in an individual directory like towhee.models.mobilenet?

jaelgu:
@gexy185 Please combine all commits into one. Make sure that 1 commit each pull request.

Issue #1149, opened 2022-04-26
[DesignProposal]: Itergate ANNS in Towhee


shiyu22:
### Background and Motivation

The Approximate Nearest Neighbor Search (ANNS) algorithm is used to search the embedding vectors efficiently, and towhee can integrate them.

### Design

There is an example for reverse image search with [Milvus](https://milvus.io/), which is an open-source vector database. I think it can be used in each ANNS lib or tool.
For inserting data:
```python
from towhee import connector

milvus_collection = connector.milvus('collection_url')                
faiss_index = connector.faiss('path/to/ann_index')
                    

with milvus_collection, faiss_index:
    (
        towhee.glob['path']('./*jpg')
            .image_decode['path', 'img']()
            .image_embedding.timm['img', 'vec1'](model_name='resnet50')
            .image_embedding.timm['img', 'vec2'](model_name='resnet101')
            .ann_insert['path', 'vec1'](milvus_collection)
            .ann_insert['path', 'vec2'](faiss_index)
    )
# batch inserting
with milvus_collection, faiss_index:
    (
        towhee.glob['path']('./*jpg')
            .image_decode['path', 'img']()
            .image_embedding.timm['img', 'vec1'](model_name='resnet50')
            .image_embedding.timm['img', 'vec2'](model_name='resnet101')
            .batch(100)
            .ann_insert['path', 'vec1'](milvus_collection)
            .ann_insert['path', 'vec2'](faiss_index)
    )
```
For searching:
```python
with milvus_collection, faiss_index:
    (
				towhee.glob['path']('query.jpg')
    				.image_decode['path', 'img']()
    				.image_embedding.timm['img', 'vec1'](model_name='resnet50')
    				.image_embedding.timm['img', 'vec2'](model_name='resnet101')
    				.ann_search['vec1', 'result1'](milvus_collection) #optional `search_kwargs` for the search parameters 
    				.ann_search['vec2', 'result2'](faiss_index)
    )
```
For inserting and searching:
```python
with milvus_collection, faiss_index:
    (
        towhee.glob['path']('./*jpg')
            .image_decode['path', 'img']()
            .image_embedding.timm['img', 'vec1'](model_name='resnet50')
            .image_embedding.timm['img', 'vec2'](model_name='resnet101')
            .ann_insert['path', 'vec1'](milvus_collection)
            .ann_insert['path', 'vec2'](faiss_index)

       faiss_index.train()
      
				towhee.glob['path']('query.jpg')
[...]

PR #1159, opened 2022-04-27
Add ANNIndex base and milvus db

shiyu22:

### DIFF
--- a/towhee/__init__.py
+++ b/towhee/__init__.py
@@ -11,13 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import os
-from typing import List, Union
+from typing import Union
 from towhee.engine import register
 from towhee.engine.factory import ops, pipeline, DEFAULT_PIPELINES, _PipelineBuilder as Build
 from towhee.hparam import param_scope
 from towhee.hparam import HyperParameter as Document
 from towhee.functional import DataCollection, State, Entity
+from towhee.connectors import Connectors as connectors
 
 from towhee.functional import glob, glob_zip
 from towhee.functional import read_csv, read_json, read_camera
@@ -37,6 +37,7 @@ __all__ = [
     'DataCollection',
     'State',
     'Entity',
+    'connectors',
     #
     'glob',
     'glob_zip',
--- a/towhee/functional/data_collection.py
+++ b/towhee/functional/data_collection.py
@@ -17,6 +17,7 @@ from random import random, sample, shuffle
 import reprlib
 
 from towhee.hparam import param_scope, dynamic_dispatch
+from towhee.functional.entity import Entity
 from towhee.functional.option import Option, Some, Empty
 from towhee.functional.mixins import AllMixins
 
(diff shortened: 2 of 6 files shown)


towhee/__init__.py
```diff
@@ -11,13 +11,13 @@
...
 # limitations under the License.
-import os
-from typing import List, Union
+from typing import Union
 from towhee.engine import register
 from towhee.engine.factory import ops, pipeline, DEFAULT_PIPELINES, _PipelineBuilder as Build
 from towhee.hparam import param_scope
 from towhee.hparam import HyperParameter as Document
 from towhee.functional import DataCollection, State, Entity
+from towhee.ann_index import Connector as connector
```
reiase, line 20: It would be better to use `connectors` as the module name.

towhee/ann_index/base.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        pass
+
+    def save(self):
+        """
+        Save the index.
+        """
+        pass
+
+    def __enter__(self):
```
reiase, line 62: can we put the code of entering and exiting here?

towhee/functional/data_collection.py
```diff
@@ -459,40 +460,66 @@ def range(*arg, **kws):
         """
         return DataCollection(range(*arg, **kws))
 
-    def batch(self, size, drop_tail=False) -> 'DataCollection':
+    def batch(self, size, drop_tail=False, raw=True):
```
reiase, line 463: can we just put the entities into a list here, and don't deal with the fields.

reiase (changes requested):

towhee/functional/data_collection.py
shiyu22: I have tried it but not work. If we yield a list of entities will got:
```bash
~/workspace/git/towhee/towhee/engine/factory.py in <listcomp>(.0)
     73     def __apply__(self, *arg, **kws):
     74         if isinstance(self._index[0], tuple): # Multi inputs.
---> 75             args = [getattr(arg[0], x) for x in self._index[0]]
     76         else: # Single input.
     77             args = [getattr(arg[0], self._index[0])]

AttributeError: 'list' object has no attribute 'path'
```
and if yield DataCollection directly, the args in op is like `(<towhee.hparam.hyperparameter.wrapper object at 0x7f95031df460>, <towhee.hparam.hyperparameter.wrapper object at 0x7f95031df520>)`.

And the example code is follows:
```python
import towhee
from towhee import connector

milvus_collection = connector.milvus(uri='tcp://localhost:19530/reverse_image_search')            
                    

with milvus_collection:
    (
        towhee.glob['path']('/Users/chenshiyu/workspace/data/pic/test/*.jpg')
              .image_decode['path', 'img']()
              .image_embedding.timm['img', 'vec'](model_name='resnet50')
              .runas_op['path', 'path'](func=lambda path: abs(hash(path)) % (10 ** 8))
              .batch(3)
              .ann_insert[('path', 'vec'), 'res'](ann_index=milvus_collection)
              .run()
    )
```

towhee/ann_index/base.py
shiyu22: What do you mean, such as `load` with `__exit__`? But I think each index will have different action with `__enter__` and `__exit__`.

towhee/__init__.py
shiyu22: get it.

towhee/functional/data_collection.py
reiase: @GuoRentong We run into this problem angin, the behavior of batch  has not been defined for entities. I'll open another issue for this problem.

reiase (changes requested):

reiase (approved):

shiyu22:
#1149

PR #1171, opened 2022-04-29
Removing conditional returns and non factory calls

filip-halt:
All internal DC creation limited to _factory calls, all functions either return a datacollection or don't return a collection.

### DIFF
--- a/towhee/functional/data_collection.py
+++ b/towhee/functional/data_collection.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from typing import Any, Iterable, Iterator, Callable
-from random import random, sample, shuffle
+import random
 import reprlib
 
 from towhee.hparam import param_scope, dynamic_dispatch
@@ -102,19 +102,12 @@ class DataCollection(Iterable, AllMixins):
             return (x[1] for x in self._iterable.iterrows())
         return iter(self._iterable)
 
-    def stream(arg):  # pylint: disable=no-self-argument
+    def stream(self):
         """
         Create a stream data collection.
 
         Examples:
-
-        1. Create a streamed data collection
-
-        >>> dc = DataCollection.stream([0,1,2,3,4])
-        >>> dc.is_stream
-        True
-
-        2. Convert a data collection to streamed version
+        1. Convert a data collection to streamed version
 
         >>> dc = DataCollection([0, 1, 2, 3, 4])
         >>> dc.is_stream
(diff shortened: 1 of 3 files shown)


towhee/functional/data_collection.py
```diff
@@ -512,14 +493,14 @@ def inner():
                     if raw:
                         yield buff
                     else:
-                        yield DataCollection.unstream(buff if isinstance(buff, list) else [buff])
+                        yield buff if isinstance(buff, list) else [buff]
```
reiase, line 496: we expect the user can also use DC for the batch. For example, apply a towhee operator.

reiase (changes requested):

reiase:
I think it's better to make DC immutable, making it safe for multi-thread and distributed execution.

towhee/functional/data_collection.py
filip-halt: When creating the dag, each batch would be considered its own node if we make them DCs. Wouldnt it be better to just return the batches as a list, then act on those lists? Or are you expecting people to use batch to split the data collection into something like train, validate, test and perform further actions on those batches.

filip-halt: instead of returning a dc[dc, dc, dc, ...] we return a dc[list, list, list, ..].

towhee/functional/data_collection.py
```diff
@@ -125,26 +118,19 @@ def stream(arg):  # pylint: disable=no-self-argument
...
-            if self.is_stream:
-                return self
-            return DataCollection.stream(self._iterable)
-
-        iterable = arg
-        if not isinstance(iterable, Iterator):
-            iterable = iter(iterable)
-        return DataCollection(iterable)
+        if not self.is_stream:
+            self._iterable = iter(self._iterable)
```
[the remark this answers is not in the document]
reiase: It's better to create a new DataCollection like this:
```python
return DataCollection(iter(self))
```

towhee/functional/data_collection.py
```diff
@@ -157,17 +143,9 @@ def unstream(arg):  # pylint: disable=no-self-argument
...
-            if not self.is_stream:
-                return self
-            return DataCollection.unstream(self._iterable)
-
-        iterable = arg
-        if isinstance(iterable, Iterator):
-            iterable = list(iterable)
-        return DataCollection(iterable)
+        if self.is_stream:
+            self._iterable = list(self._iterable)
```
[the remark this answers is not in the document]
reiase:
```python
return DataCollection(list(self))
```

towhee/functional/data_collection.py
```diff
@@ -199,21 +175,30 @@ def is_stream(self):
...
-                hp().data_collection.parent = self
-                return creator(*arg, **kws)
+        Args:
+            iterable: An iterable object, the data being stored in the DC
+            parent_stream: Whether to copy the parents format (streamed vs unstreamed)
+
+        Returns:
+            DataCollection: DataCollection encapsulating the iterable.
+        """
+        if parent_stream is True:
```
reiase, line 191: Do we have to pass this parameter `parent_stream` on every _factory call?

In most cases, the _factoru receives a generator produced by mapping an inner function to the iterable. In such actions, DC always returns a stream child with the default `parent_stream `, but we expect an unstreamed one if the parent is unstreamed.

reiase: And there lacks a test case for checking this behavoir, stream in stream out, unstream in unstream out.

reiase (changes requested):

reiase (approved):

PR #1210, opened 2022-05-09
Add Faiss Search

junjiejiangjjj:


fzliu:
This assumes that the user has already installed `faiss`. Wouldn't it be better to package all of our ANN engines as hub operators?

junjiejiangjjj:
> This assumes that the user has already installed `faiss`. Wouldn't it be better to package all of our ANN engines as hub operators?

The anns.insert is a reduce operator, but our engine can not support reduce. We decide impl the faiss op after framework supporting reduce op. I will close this pr.

PR #1225, opened 2022-05-11
Spilt DataFrame from DataCollection to deal with Entity-based DC separately

Chiiizzzy:

### DIFF
--- a/towhee/functional/mixins/dataframe.py
+++ b/towhee/functional/mixins/dataframe.py
@@ -21,7 +21,7 @@ from towhee.hparam import dynamic_dispatch, param_scope
 # pylint: disable=protected-access
 
 
-class EntityMixin:
+class DataFrameMixin:
     """
     Mixin to help deal with Entity.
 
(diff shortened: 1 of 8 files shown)


towhee/functional/mixins/dataframe.py
```diff
@@ -30,15 +30,15 @@ class EntityMixin:
     1. define an operator with `register` decorator
 
     >>> from towhee import register
-    >>> from towhee import DataCollection
+    >>> from towhee import DataFrame
```
filip-halt, line 33: I think we have to be careful here. DataFrame already exists for the original engine. This could lead to incorrect imports. I think we either should use a different name or use absolute imports of the correct dataframe.

filip-halt (commented):
Having two DataFrame

towhee/functional/mixins/dataframe.py
Chiiizzzy: Yeah, make sense, I'll discuss with @reiase

GuoRentong: > I think we have to be careful here. DataFrame already exists for the original engine. This could lead to incorrect imports. I think we either should use a different name or use absolute imports of the correct dataframe.

This DataFrame will be as a part of Towhee API. And the original one is part of internal implementation. We will rename the original one in the next round of engine refactor.

PR #1240, opened 2022-05-12
update example

reiase:


shiyu22:
@reiase Can you update the example to the latest version?

PR #1251, opened 2022-05-16
Create col_table and convert_entity function to partly support col-based storage

Chiiizzzy:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ tqdm>=4.59.0
 ruamel.yaml<=0.16.6
 tabulate
 numpy
+pyarrow
--- a/towhee/functional/mixins/column.py
+++ b/towhee/functional/mixins/column.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from enum import Flag, auto
+
 import pyarrow as pa
 
 
@@ -18,7 +20,11 @@ class ColumnMixin:
     """
     Mixins to support column-based storage.
     """
-    def col_table(self):
+    class ModeFlag(Flag):
+        ROWBASEDFLAG = auto()
+        COLBASEDFLAG = auto()
+
+    def _create_col_table(self):
         """
         Create a column-based table.
 
@@ -27,7 +33,7 @@ class ColumnMixin:
         >>> from towhee import Entity, DataFrame
         >>> e = [Entity(a=a, b=b) for a,b in zip(['abc', 'def', 'ghi'], [1,2,3])]
         >>> df = DataFrame(e)
-        >>> table = df.col_table()
+        >>> table = df._create_col_table()
         >>> table
         pyarrow.Table
         a: string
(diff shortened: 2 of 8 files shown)


towhee/functional/mixins/column.py
```diff
@@ -0,0 +1,96 @@
...
+        ----
+        a: [["abc","def","ghi"]]
+        b: [[1,2,3]]
+        """
+        header = [*self._iterable[0].__dict__]
+        cols = [[] for _ in header]
+
+        # cols = [[getattr(entity, col) for entity in self._iterable] for col in header]
+        def inner(entity):
+            nonlocal cols, table
```
filip-halt, line 44: unused and undeclared `table`, should it be `header`?

towhee/functional/mixins/column.py
```diff
@@ -0,0 +1,96 @@
...
+        >>> table = df.col_table()
+        >>> table
+        pyarrow.Table
+        a: string
+        b: int64
+        ----
+        a: [["abc","def","ghi"]]
+        b: [[1,2,3]]
+        """
+        header = [*self._iterable[0].__dict__]
```
filip-halt, line 39: Cant subscript _iterable if streamed.

towhee/functional/mixins/column.py
```diff
@@ -0,0 +1,96 @@
...
+        cols = [[] for _ in header]
+
+        # cols = [[getattr(entity, col) for entity in self._iterable] for col in header]
+        def inner(entity):
+            nonlocal cols, table
+            for col, name in zip(cols, header):
+                col.append(getattr(entity, name))
+
+        any(map(inner, self._iterable))
```
filip-halt, line 48: whats the point of using any() here?

towhee/functional/mixins/column.py
```diff
@@ -0,0 +1,96 @@
...
+        table = pa.table(cols, names=header)
```
filip-halt, line 49: From the pyarrow api [notes](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html):
`Do not call this class’s constructor directly, use one of the from_* methods instead.`

towhee/functional/mixins/column.py
```diff
@@ -0,0 +1,96 @@
...
+        >>> e = [Entity(a=a, b=b) for a,b in zip(['abc', 'def', 'ghi'], [1,2,3])]
+        >>> df = DataFrame(e)
+        >>> df
+        [<Entity dict_keys(['a', 'b'])>, <Entity dict_keys(['a', 'b'])>, <Entity dict_keys(['a', 'b'])>]
+        >>> col_table = df.col_table()
+        >>> df.convert_entity(col_table)
+        >>> df.to_list()
+        [<Entity dict_keys(['offset', 'data'])>, <Entity dict_keys(['offset', 'data'])>, <Entity dict_keys(['offset', 'data'])>]
+        """
+        attrs = [*self._iterable[0].__dict__]
```
filip-halt, line 69: Cant subscript _iterable if streamed.

towhee/functional/mixins/column.py
Chiiizzzy: I forgot to delete this from previous version.

towhee/functional/mixins/column.py
Chiiizzzy: `inner()` does not return any value, so we need an `any()` or `all()` to execute `inner()`, otherwise `inner()` won't be executed.

towhee/functional/mixins/column.py
Chiiizzzy: Fixed

towhee/functional/mixins/column.py
Chiiizzzy: Cool, fixed

towhee/functional/mixins/column.py
Chiiizzzy: Fixed

towhee/functional/mixins/column.py
filip-halt: This is a dangerous action, if we accidentally change inner at some point, and introduce the possibility of returning true, the map wont be run completely. I think for this case using map is pointless, as we are generating useless data. Instead I would just use a regular for loop:
```python
for x in self._iterable:
   inner(x)
```

https://stackoverflow.com/questions/1080026/is-there-a-map-without-result-in-python/1080300#1080300
https://stackoverflow.com/questions/47722712/use-map-for-functions-that-does-not-return-a-value

Chiiizzzy: But I think we use `map()` here to realize stream computing. If we use for loop, we need to load the whole `self._iterable` which might cause OOM

filip-halt: if the self._iterator is a list it will already be loaded into memory, if its an iterator then it will be iterated over (stream based) I believe.

requirements.txt
```diff
@@ -3,3 +3,4 @@ tqdm>=4.59.0
 ruamel.yaml<=0.16.6
 tabulate
 numpy
+pyarrow
```
reiase, line 6: reconsider the dependency of yarrow

PR #1269, opened 2022-05-19
auto_download packages

filip-halt:

### DIFF
--- a/towhee/utils/dependency_control.py
+++ b/towhee/utils/dependency_control.py
@@ -1,0 +1,39 @@
+# Copyright 2022 Zilliz. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


towhee/utils/dependency_control.py
```diff
@@ -0,0 +1,39 @@
...
+    to True then the package will be automatically installed.
+    """
+    if os.getenv('TOWHEE_WORKER', 'False') == 'True' or get_yes_no(f'Do you want to install {package}?'):
+        try:
+            subprocess.check_call([sys.executable, '-m', 'pip', 'install', package])
+            print(f'{package} installed successfully!')
+        except subprocess.CalledProcessError:
+            print(f'Ran into error installing {package}.')
+
+def get_yes_no(question): # pragma: no cover
```
reiase, line 31: I think we should also ask the user whether to install the package into user directory, see `pip install --user`

towhee/utils/dependency_control.py
```diff
@@ -0,0 +1,39 @@
...
+    while True:
+        answer = input(question + ' [y/n]: ').lower()
```
reiase, line 33: question + '[y/u/n]', and u for install to user directory

reiase (changes requested):

towhee/utils/dependency_control.py
filip-halt: I was originally going to do this but then I didnt because using -u will mess up if the user is using a virtual_env

reiase: Let's move this discussion to another issue, and I will merge this PR

reiase (approved):

PR #1293, opened 2022-05-26
feat: enhance save-images api

soulteary:
this PR contains:

- Support for specifying image types
- Support to keep the name of the input image

thanks @shiyu22

### DIFF
--- a/towhee/hub/builtin/operators/computer_vision.py
+++ b/towhee/hub/builtin/operators/computer_vision.py
@@ -227,14 +227,21 @@ class save_image:
     >>> (
     ...     towhee.dc['path'](['https://github.com/towhee-io/towhee/raw/main/towhee_logo.png'])
     ...           .image_decode['path', 'img']()
-    ...           .save_image['img','path_new'](dir='temp/pic')
+    ...           .save_image['img','path_new'](dir='temp/pic', format="jpg", keep_original_filename = False)
     ...           .to_list()
     ... )
     [<Entity dict_keys(['path', 'img', 'path_new'])>]
     """
 
-    def __init__(self, dir: str):
+    def __init__(self, dir: str, format: str, keep_original_filename: bool):
         self._dir = dir
+        # specify image format
+        if format == "":
+            self._format = "jpg"
+        else:
+            self._format = format
+        # save original file name
+        self._keep_original_filename = keep_original_filename
         if not Path(self._dir).exists():
             Path(self._dir).mkdir(parents=True)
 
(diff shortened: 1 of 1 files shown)


towhee/hub/builtin/operators/computer_vision.py
```diff
@@ -227,22 +227,34 @@ class save_image:
...
     ...           .image_decode['path', 'img']()
-    ...           .save_image['img','path_new'](dir='temp/pic')
+    ...           .save_image['img','path_new'](dir='temp/pic', format="jpg", keep_original_filename = False)
     ...           .to_list()
     ... )
     [<Entity dict_keys(['path', 'img', 'path_new'])>]
     """
 
-    def __init__(self, dir: str):
+    def __init__(self, dir: str, format: str, keep_original_filename: bool):
```
shiyu22, line 236: I think the `keep_original_filename` parameter is unnecessary, and `format` can defaults to '.jpg', like:`def __init__(self, dir: str, format: str='.jpg'):`

towhee/hub/builtin/operators/computer_vision.py
```diff
@@ -227,22 +227,34 @@ class save_image:
...
+        self._keep_original_filename = keep_original_filename
         if not Path(self._dir).exists():
             Path(self._dir).mkdir(parents=True)
 
     def __call__(self, img):
         from towhee.utils.pil_utils import PILImage
         from towhee.utils.ndarray_utils import cv2
 
-        self._file = str(uuid.uuid4()) + '.jpg'
+        if self._keep_original_filename == True:
```
shiyu22, line 252: If we want to use the origin filename, `__call__`function need a new parameter like `ori_path` and it defaults to None, so that new filename will like:
```python
def __call__(self, img, ori_path=None):
    if ori_path:
        self._file = ori_path.split('/')[-1]
    else:
        self._file = str(uuid.uuid4()) + '.' + self._format
```

And when we call save_image, it will support two parameters:
```python
dc = towhee.glob['path']('.test/*.JPEG') \
      .image_decode['path', 'img'() \
      .img2img_translation.cartoongan['img', 'new_img'](model_name = 'Hayao', device='cuda') \
      .save_image[('new_img', 'path'), 'new_path'](dir='tmp') \
      .to_list()
```

towhee/hub/builtin/operators/computer_vision.py
```diff
@@ -227,22 +227,31 @@ class save_image:
...
     ...           .to_list()
     ... )
     [<Entity dict_keys(['path', 'img', 'path_new'])>]
     """
 
-    def __init__(self, dir: str):
+    def __init__(self, dir: str, format: str='.jpg'):
         self._dir = dir
+        # specify image format
+        if format == "":
```
shiyu22, line 239: There is no need to judge.

shiyu22:
Can you update the docstring with the new parameter and more examples with the orig_path?

PR #1314, opened 2022-06-02
AudioFrame add layout info

junjiejiangjjj:

### DIFF
--- a/towhee/types/audio_frame.py
+++ b/towhee/types/audio_frame.py
@@ -24,12 +24,13 @@ class AudioFrame(np.ndarray):
             The audio sample rate in Hz.
     """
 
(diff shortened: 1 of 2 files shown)


towhee/types/audio_frame.py
```diff
@@ -24,12 +24,13 @@ class AudioFrame(np.ndarray):
             The audio sample rate in Hz.
     """
```
reiase, line 25: Can we add some test cases in docstring so that the code blocks will appear in the API docs?

junjiejiangjjj: The AudiaoFrames are created by audio_decoder.ffmpeg, we will add an example in decoder op's README.

Issue #1275, opened 2022-05-20
[Enhancement]: Some doctests are not included in our workflow


Chiiizzzy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What would you like to be added?

There are some doctests not included in our workflow:
- functional.__init__.py

filip-halt:
Is someone working on this?

PR #1330, opened 2022-06-07
fix 1329

### LINKED ISSUE #1329: [Bug]: .dc() fails to work with more than one column
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

import towhee
dc = towhee.dc\(\["num","val"\](\[\[1,"one"\],\[2,"two"\]\]\)

Throws an error. 

Fix
return DataFrame(iterable).map(lambda x: Entity(**{index: x})) -------> return DataFrame(iterable).map(lambda x: Entity(**dict(zip(index, x))))

### Expected Behavior

To get a dc with two columns.

### Steps To Reproduce

_No response_

### Environment

```markdown
- Towhee version(e.g. v0.1.3 or 8b23a93):
- OS(Ubuntu or CentOS):
- CPU/Memory:
- GPU:
- Others:
```

### Anything else?

_No response_

filip-halt:
fixes #1329 

### DIFF
--- a/towhee/functional/__init__.py
+++ b/towhee/functional/__init__.py
@@ -167,9 +167,13 @@ def dc(iterable):
 
     >>> towhee.dc['column']([0, 1, 2]).to_list()
     [<Entity dict_keys(['column'])>, <Entity dict_keys(['column'])>, <Entity dict_keys(['column'])>]
+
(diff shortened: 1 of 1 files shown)


reiase:
The PR may break some of the doctest, some of the doctest are not running as reported in #1275.

reiase:
This issue is also fixed in #1331.

Issue #1270, opened 2022-05-20
[Bug]: to_milvus unstream is unnecessary and causes OOM


filip-halt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

to_milvus unstreams, causing a memory bottleneck. If inserting 100,000 images into milvus, the pipeline should be streamed. 

### Environment

```markdown
- Towhee version(e.g. v0.1.3 or 8b23a93):
- OS(Ubuntu or CentOS):
- CPU/Memory:
- GPU:
- Others:
```

shiyu22:
`to_milvus` is unstream by default, in order to support inserting different data into different collections:
```bash
(
	towhee.glob['path']('./train/*/*.JPEG')
		.image_decode['path', 'img']()
		.image_embedding.timm['img', 'vec'](model_name='resnet50')
                 .image_embedding.timm['img', 'vec'](model_name='resnet101')
		.runas_op['path', 'label'](func = lambda path: path.split('/')[-2])
		.runas_op['path', 'path'](func=lambda path: abs(hash(path)) % (10 ** 8))
		.to_milvus['path', 'vec'](collection=collection_res50, batch=100)
                .to_milvus['path', 'vec'](collection=collection_res101, batch=100)
)
```

And anyone can add unstream=False to change the mode when using to_milvus, like `.to_milvus['id', 'vec'](collection=collection, batch=100, unstream=False)`.

shiyu22:
The default `unstream` mode may work for demos, how do you think? @filip-halt @reiase

reiase:
> The default `unstream` mode may work for demos, how do you think? @filip-halt @reiase

can we fix that by returning another DC in `to_milvus `

PR #1331, opened 2022-06-08
fix #1275

### LINKED ISSUE #1275: [Enhancement]: Some doctests are not included in our workflow

reiase:

doctest finder does not support global variables, we can't directly apply the decorator to the function if we want both sphinx docs and doctest.

### DIFF
--- a/towhee/doc/source/data_collection/get_started.rst
+++ b/towhee/doc/source/data_collection/get_started.rst
@@ -2,5 +2,5 @@
 Get Started
 ===========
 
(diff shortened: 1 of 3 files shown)


reiase:
close #1270 
close #1329

PR #1359, opened 2022-06-10
add load pretrained function to video_swin_transformer

jinlingxu06:


jaelgu (changes requested):
1. Change print out messages back to log
2. Change api to better support operator:
   - forward_features
   - head
   - forward (forward_features + head)

Issue #1342, opened 2022-06-09
[DesignProposal]: operator workload refactoring


reiase:
### Background and Motivation

1. Towhee is trying to support various different execution backend, for example, vectorized execution, execution with native code generation, and execution on ray;
2. The entrance of execution is `_OperatorLazyWrapper.__call__`, which has become too complex for further maintenance;
3. wrapper is not good for performance and code generation;

### Design

1. Decompose the class `_OperatorLazyWrapper ` with the mixin design pattern; #1343 
2. provide a `compile()` which will generate an optimized and jit-friendly callable for the input;

PR #1360, opened 2022-06-10
[WIP] Support JIT with ops

shiyu22:

#1342 

```python
import numpy
import time
from towhee import register, ops


@register(name='inner_distance')
def inner_distance(query, data):
    dists = []
    for vec in data:
        dist = 0
        for i in range(len(vec)):
            dist += vec[i] * query[i]
        dists.append(dist)
    return dists


if __name__ == '__main__':
    data = numpy.random.random((10000, 128))
    query = numpy.random.random(128)

    op = ops.inner_distance()
    time1 = time.time()
    dists = op(query, data)
    print(time.time() - time1)

    time1 = time.time()
    dists = op(query, data)
    print(time.time() - time1)
```

0.8242287635803223
0.0020759105682373047


shiyu22:
#1342

PR #1391, opened 2022-06-15
modify frozen in time model to adapt bridgeformer

jinlingxu06:


jaelgu (changes requested):
@jinlingxu06 Change print out message to log

PR #1401, opened 2022-06-16
Add UT for mimo

Chiiizzzy:

### DIFF
--- a/towhee/engine/execution/vectorized_execution.py
+++ b/towhee/engine/execution/vectorized_execution.py
@@ -12,10 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# import numpy as np
-
-# from towhee.types.tensor_array import TensorArray
-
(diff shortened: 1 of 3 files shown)


towhee/engine/execution/vectorized_execution.py
```diff
@@ -28,9 +25,17 @@ def __vcall__(self, *arg, **kws):
         if hasattr(self._op, '__vcall__'):
             return self._op.__vcall__(*arg, **kws)
         elif len(arg) == 1:
-            return [self._op(x) for x in arg[0]]
+            res = [self._op(x) for x in arg[0]]
+            try:
```
reiase, line 29: why we need a try-catch here?

Chiiizzzy: For multi output cases, we need to turn [(o1, o2), (o3, o4) ..] to ([o1, o3, ...], [o2, o4, ...])

PR #1402, opened 2022-06-16
add create_model of movinet

gexy185:
Signed-off-by: gexy5 <xinyu.ge@zilliz.com>


jaelgu (changes requested):
@gexy185 No value is assigned to `device` in `create_model`

PR #1408, opened 2022-06-16
modify clip to adapt bridgeformer

jinlingxu06:


jaelgu (changes requested):
Remove
```
if self.is_bridge_former:
            return text_features, image_features
```
from `clip.forward`.
Duplicate of clip.encode_text & clip.encode_image

PR #1436, opened 2022-06-21
modify timesformer to adapt svt

jinlingxu06:


jaelgu (changes requested):
Add `model_name` in svt ut to load configs & test model with svt model name.

PR #1448, opened 2022-06-22
fix occasional failure of UT

junjiejiangjjj:


tests/unittests/engine/operator_runner/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def run(runner):
+    runner.process()
+
+
+def start_runner(runner):
+    t = threading.Thread(target=run, args=(runner, ))
+    t.start()
+    time.sleep(0.01) # make sure runner started.
```
reiase, line 12: Why we need a magic number here?

tests/unittests/engine/operator_runner/test_concat_runner.py
```diff
@@ -74,8 +71,7 @@ def test_concat_runner(self):
             writer
         )
         runner.set_op(ConcatOperator('row'))
-        t = threading.Thread(target=run, args=(runner, ))
-        t.start()
+        t = start_runner(runner)
```
reiase, line 74: I think we need a stable checker for the runner.

reiase (changes requested):

tests/unittests/engine/operator_runner/__init__.py
junjiejiangjjj: Just make sure the runner is running.

tests/unittests/engine/operator_runner/test_concat_runner.py
junjiejiangjjj: Only for UT, because the ut will check the runner status after the runner start.

PR #1506, opened 2022-07-04
add coformer

gexy185:
Signed-off-by: gexy5 <xinyu.ge@zilliz.com>


jaelgu (changes requested):
@gexy185 
- Move `position_encoding.py` to `towhee.models.layers`
- Try to remove `process_image` after comparing with `towhee.models.utils.video_transforms`

PR #1511, opened 2022-07-05
add trtion_client

zhujiming:


serve/triton/client.py
```diff
@@ -0,0 +1,179 @@
...
+			self.client = httpclient.InferenceServerClient(url)
+
+	def __del__(self):
+		if self.protocol == 'grpc':
+			self.stop_stream()
+
+	def get_model_name(self):
+		return self._model_name
+
+	def async_set(self, iterator):
```
junjiejiangjjj, line 57: If the input is an iterator, `len(iterator)` is error

serve/triton/client.py
```diff
@@ -0,0 +1,179 @@
...
+import sys
+import numpy as np
+import tritonclient.grpc as grpcclient
+import tritonclient.http as httpclient
+from tritonclient.utils import InferenceServerException
+from functools import partial
+
+model_name = 'pipeline'
+
+if sys.version_info >= (3, 0):
```
junjiejiangjjj, line 24: `towhee` support python>=3.6, so this code is not required.

serve/triton/client.py
```diff
@@ -0,0 +1,179 @@
...
+			(results, error) = self.user_data._completed_requests.get()
+			processed_count += 1
+			if error is not None:
+				sys.exit(1)
+			if processed_count <= len(iterator):
+				responses.append(results)
+		
+		return self._solve_responses(outputs, responses)
+
+	def stream(self, iterator):
```
junjiejiangjjj, line 80: what is the difference between `async_set ` and `stream`, if the protocol is grpc, all of them will call `start_stream`

serve/triton/client.py
zhujiming: async_set and stream function must use length to column request times , so these 'iterator''s type is list. Input must list

serve/triton/client.py
```diff
@@ -0,0 +1,179 @@
...
+
+		sent_count = 0
+		responses = []
+		try:
+			for i in range(len(iterator)):
+				sent_count += 1
+				self.client.async_stream_infer(self.get_model_name(), inputs, request_id=str(sent_count), outputs=outputs)
+		except InferenceServerException as e:
+			self.stop_stream()
+			sys.exit(1)
```
junjiejiangjjj, line 93: We couldn't exit the process after call service failed.

serve/triton/client.py
```diff
@@ -0,0 +1,179 @@
...
+# Callback function used for async_stream_infer()
+def completion_callback(user_data, result, error):
+	# passing error raise and handling out
+	user_data._completed_requests.put((result, error))
+
+class Client():
+	def __init__(self, url, model_name=model_name, protocol='grpc'):
+		self._model_name = model_name
+		self.user_data = UserData()
+		if protocol == 'grpc':
```
junjiejiangjjj, line 38: the code `if protocal==grpc else ...` appears in many place. 
It seems we have `httpclient`, `streamclient` and `asycclient`

PR #1513, opened 2022-07-05
add cvnet model

jinlingxu06:


jaelgu (changes requested):
@jinlingxu06 

- cvnet_utils: Remove useless functions in `Geometry`

PR #1522, opened 2022-07-07
modify swin_transformer to adapt swin_transformerV2

jinlingxu06:


jaelgu (changes requested):
@jinlingxu06 
- Remove or comment `main` function call (increase coverage)
- init_vit_weights.py: std as parameter, defaults to 1e^-6 OR .02

PR #1531, opened 2022-07-08
modify utils to adapt swin_transformer_v2

jinlingxu06:


jaelgu (changes requested):
@jinlingxu06 Add ut for init_vit_weights.py (test with small dummy weights dict & module)

PR #1551, opened 2022-07-12
Clean up cache structure

Chiiizzzy:


shiyu22:
@Chiiizzzy When running op or pipeline with `towhee.ops`, it will download the repo twice to different directories because `towhee.ops` can also run with pipeline.

PR #1576, opened 2022-07-18
Fix issue 1572

zhujiming:


test_requirements.txt
```diff
@@ -24,6 +24,4 @@ transformers>=4.19.2
 einops>=0.4.1
 timm>=0.4.12
 faiss-cpu
-numba
-tritonclient
-geventhttpclient
\ No newline at end of file
+numba
```
reiase, line 27: tritonclient and geventhttpclient should be added to test_requirements.txt

zhujiming: fixed

PR #1594, opened 2022-07-20
add op_config to dag

zhujiming:


towhee/engine/factory.py
```diff
@@ -101,6 +101,10 @@ def get_op(self):
         self.__check_init__()
         return self._op
 
+    def get_op_config(self):
```
reiase, line 104: please use `@property` instead of `get_xxx`

zhujiming: fixed

reiase (approved):

PR #1630, opened 2022-07-26
Add README_CN.md

songxianj:
Signed-off-by: JIANG Songxian [songxian.jiang@zilliz.com](mailto:songxian.jiang@zilliz.com).

One thing I'm not sure about: would you double-check the translation of "Pythonic API"? This phrase does not have a generally accepted Chinese equivalent, and could be simply kept as is.

### DIFF
--- a/README_CN.md
+++ b/README_CN.md
@@ -1,0 +1,166 @@
+&nbsp;
+
+<p align="center">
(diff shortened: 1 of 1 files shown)


fzliu:
Dup #1644

PR #1682, opened 2022-08-10
support schema

zhujiming:


filip-halt:
You will probably need to squash all those commits.

PR #1697, opened 2022-08-15
add ut in docker

zhujiming:


.github/workflows/codecov.yml
```diff
@@ -14,36 +14,19 @@ on:
...
-        conda activate towhee-ut
-        conda install pytorch torchvision torchaudio cudatoolkit=11.3 -c pytorch
-        conda run -n towhee-ut pip install coverage pytest pytest-xdist
-        conda run -n towhee-ut pip install -r test_requirements.txt
-        conda run -n towhee-ut coverage erase
-        conda run -n towhee-ut coverage run -m pytest
-    - name: Generate xml report
-      run: |
+        rm -rf towhee
+        git clone https://github.com/towhee-io/towhee.git
```
reiase, line 23: I think it is better to use `actions/checkout` instead of using the git command line directly. The action should be able to checkout the right branch for testing.

reiase (changes requested):

.github/workflows/codecov.yml
Chiiizzzy: Also, by clone repo every time, we are running ut on main branch, so the PR is not tested, we need to change the way we load tests dir.

PR #1714, opened 2022-08-16
Documentation/mixins comments

filip-halt:


towhee/functional/mixins/serve.py
```diff
@@ -21,8 +21,10 @@
 
 
 class _APIWrapper:
-    """
-    API Wrapper
+    """API Wrapper
```
reiase, line 24: Can we hide the private classes and functions from the document? A function/class is considered to be private and should only be used in the same file it is defined if its name is started with "_".

filip-halt: i think we can

PR #1737, opened 2022-08-18
Allow users to selectively use MIRROR when building docker images

soulteary:
When the current user builds the Towhee image, they is forced to use the university software source in China.

Maybe it makes more sense to make this optional.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,16 @@ ARG PYTHON_VERSION=3.8
 # base image, build command: 
 # docker build --platform x86_64 --target towhee-base  -t towhee/towhee-base:latest .
 FROM ${BASE_IMAGE} as towhee-base
-RUN sed -i "s@http://.*archive.ubuntu.com@http://mirrors.tuna.tsinghua.edu.cn@g" /etc/apt/sources.list
-RUN sed -i "s@http://.*security.ubuntu.com@http://mirrors.tuna.tsinghua.edu.cn@g" /etc/apt/sources.list
+
+# Speed ​​up for Chinese users, only if user specified `USE_MIRROR=true`
+# In addition, use the default software source of the base image and language
+ARG USE_MIRROR=false
+ENV USE_MIRROR=${USE_MIRROR}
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,14 +4,16 @@ ARG PYTHON_VERSION=3.8
...
 # docker build --platform x86_64 --target towhee-base  -t towhee/towhee-base:latest .
 FROM ${BASE_IMAGE} as towhee-base
-RUN sed -i "s@http://.*archive.ubuntu.com@http://mirrors.tuna.tsinghua.edu.cn@g" /etc/apt/sources.list
-RUN sed -i "s@http://.*security.ubuntu.com@http://mirrors.tuna.tsinghua.edu.cn@g" /etc/apt/sources.list
+
+# Speed ​​up for Chinese users, only if user specified `USE_MIRROR=true`
+# In addition, use the default software source of the base image and language
+ARG USE_MIRROR=false
+ENV USE_MIRROR=${USE_MIRROR}
+RUN if [ "$USE_MIRROR" = "true" ]; then sed -i -e "s/archive.ubuntu.com/mirrors.tuna.tsinghua.edu.cn/" /etc/apt/sources.list && sed -i -e "s/security.ubuntu.com/mirrors.tuna.tsinghua.edu.cn/" /etc/apt/sources.list; fi
```
GuoRentong, line 12: we should add an argument 'mirror=', rather than giving a fixed mirro address

GuoRentong (changes requested):
please check the comments i left

Dockerfile
GuoRentong: and a list of mirro should be given in user docs

soulteary: Received suggestions, I think your idea is more suitable for users. @GuoRentong 

I will improve this commit.

soulteary:
According to @GuoRentong  suggestion, I added two build parameters, which are used to control whether MIRROR needs to be enabled as a whole during the image building process, and whether to set specific image sources according to their own needs.

```bash
docker build --build-arg=USE_MIRROR=true --build-arg=UBUNTU_MIRROR=mirror.lzu.edu.cn -t towhee . 
```

If the user wants to use mirror, you should first configure the `use_mirror = true`, and then it will be able to enable the Mirror function at multiple levels.

If the user wants to specify a specific mirror to get the highest download speed, then you can specify the parameter of `ubuntu_mirror` to switch the specific

soulteary:
https://github.com/towhee-io/towhee/issues/1738

Sorry, I forgot to send a associated Issue address. I hope there are more suggestions or guidance, thank you.

PR #1740, opened 2022-08-18
Support flatten with schema

Chiiizzzy:


towhee/functional/mixins/data_processing.py
```diff
@@ -249,24 +249,44 @@ def inner():
         return self._factory(inner())
 
     @register_dag
-    def flatten(self) -> 'DataCollection':
+    def flatten(self, index: str = None) -> 'DataCollection':
```
reiase, line 252: the field to flatten should be passed by `[]`, for example:
```
dc.flatten['c']()
```

PR #1745, opened 2022-08-18
Add depth 1 when git clone hub repos

jaelgu:


reiase:
There will be problem if we try to pull updates from the hub.

jaelgu:
Git will be replaced by pip later. This issue will not affect main functions. Ignore for now.

PR #1815, opened 2022-08-22
add dag control plane doc

zhujiming:


filip-halt:
Honestly we can probably remove control plane and integrate it into the mixin. Was originally made as the design docs had a control plane that would be doing more, but we moved away from that design.

PR #2045, opened 2022-09-02
Support jit in triton

shiyu22:


towhee/serve/triton/dockerfiles/DockerfileCuda114
```diff
@@ -27,6 +27,10 @@ RUN apt-get install git-lfs && \
 
 WORKDIR /workspace
 
+RUN git clone -b nebullvm https://github.com/towhee-io/towhee-compiler.git
```
junjiejiangjjj, line 30: Why not build a pip package, and use pip to install it.

shiyu22: @reiase Can we release a version base on this branch?

PR #2084, opened 2022-09-19
Add faiss insert and search mixin

Chiiizzzy:


filip-halt:
Why are we creating mixins for this. Cant we use/modify https://towhee.io/tasks/detail/operator?field_name=ANN-Search&task_name=ANN-Search

PR #2086, opened 2022-09-20
[WIP] Add runtime classes

junjiejiangjjj:
Design the runtime classes


towhee/runtime/pipeline.py
```diff
@@ -0,0 +1,45 @@
...
+from towhee.runtime.graph import Graph
+
+
+class DataCollection:
+    '''Mock class
+    '''
+    pass
+
+
+class Pipeline:
```
filip-halt, line 27: can we rename this to something else. The forward facing class is also called pipeline which might get confusing.

PR #2115, opened 2022-09-23
Add runtime data_queue

junjiejiangjjj:
col-based storage with row-based interface

### DIFF
--- a/towhee/runtime/data_queue.py
+++ b/towhee/runtime/data_queue.py
@@ -1,0 +1,162 @@
+# Copyright 2021 Zilliz. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import List, Tuple
+import threading
+from enum import Enum, auto
+
+from collections import deque, namedtuple
+
+
+class DataQueue:
+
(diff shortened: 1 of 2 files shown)


towhee/runtime/data_queue.py
```diff
@@ -0,0 +1,153 @@
...
+            self._not_empty.notify_all()
+
+    @property
+    def sealed(self) -> bool:
+        return self._sealed
+
+
+class ColumnType(Enum):
+    QUEUE = auto()
+    NOQUEUE = auto()
```
GuoRentong, line 104: ARRAY?

towhee/runtime/data_queue.py
```diff
@@ -0,0 +1,153 @@
...
+# limitations under the License.
+
+from typing import List, Tuple
+import threading
+from enum import Enum, auto
+
+from collections import deque, namedtuple
+
+
+class DataQueue:
```
GuoRentong, line 22: why can't we directly wrap one of the Queue types in python queue module?

GuoRentong (changes requested):

towhee/runtime/data_queue.py
junjiejiangjjj: Queue is a thread-safe wrapper of deque

towhee/runtime/data_queue.py
junjiejiangjjj: NOQUEUE   means only having one cell,  a list, an int, or others.

towhee/runtime/data_queue.py
```diff
@@ -0,0 +1,153 @@
...
+        for col_type in self._schema.col_types():
+            if col_type == ColumnType.QUEUE:
+                self._data.append(_QueueColumn())
+            else:
+                self._data.append(_NoQueueColumn())
+            
+        self._sealed = False
+        self._size = 0
+        self._lock = threading.Lock()
+        self._not_full = threading.Condition(self._lock)
```
filip-halt, line 37: by using these conditions wont we require a seperate thread for every queue?

junjiejiangjjj: yeah, each node of the graph will have a thread

PR #2195, opened 2022-10-12
Add runtime DagRepr and SchemaRepr

shiyu22:


towhee/runtime/node_repr.py
```diff
@@ -150,8 +151,8 @@ def __init__(
             op_info: OperatorRepr,
             config: Dict[str, Any],
             next_nodes: List,
-            in_edges: List = None,
-            out_edges: List = None,
+            in_edges: List = [],
```
junjiejiangjjj, line 154: Do not use mutable type as the param's default value.

PR #2209, opened 2022-10-19
Support flat_map node

Chiiizzzy:


towhee/runtime/nodes/_flat_map.py
```diff
@@ -0,0 +1,68 @@
...
+        return ret
+
+    def process_step(self) -> List[Any]:
+
+        data = self._in_ques[0].get_dict()
+        if data is None:
+            self._set_finished()
+            return True
+
+        nested_data = [data.get(key) for key in self._node_repr.inputs]
```
junjiejiangjjj, line 40: The data `FlatMap` process is not a nested_data, 
for example:
    flat_map('url', 'image', ops.video_decode.ffmpeg()),  in there, the data is url

PR #2225, opened 2022-10-25
Add ops

shiyu22:


towhee/runtime/factory.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @property
+    def init_args(self):
+        return self._init_args
+
+    @property
+    def init_kws(self):
+        return self._init_kws
+
+    def __call__(self, *args, **kws):
```
junjiejiangjjj, line 91: We should store the op object

shiyu22: Done

PR #2227, opened 2022-10-25
DataQueue return Empty

junjiejiangjjj:

### DIFF
--- a/towhee/runtime/data_queue.py
+++ b/towhee/runtime/data_queue.py
@@ -65,7 +65,7 @@ class DataQueue:
             return True
 
     def put_dict(self, inputs: Dict) -> bool:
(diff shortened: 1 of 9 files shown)


towhee/runtime/data_queue.py
```diff
@@ -284,7 +301,7 @@ def __init__(self):
         self._data = None
 
     def put(self, data):
-        if isinstance(data, _Empty):
+        if data is Empty():
```
GuoRentong, line 304: This line checks if `data` and `Empty()` are the same object. Can we asume that all the instance of Empty are identical? 
I think we should avoid this kind of comparision and provide a util function like is_empty(value). Be careful about its implementation.

junjiejiangjjj: Empty is a singleton, to avoid creating too many objs.

PR #2231, opened 2022-10-27
Add Pipeline and OpAction

shiyu22:


towhee/runtime/pipeline.py
```diff
@@ -0,0 +1,394 @@
...
+                'type': 'concat',
+                'param': None
+            },
+            'next_nodes': [],
+        }
+        dag_dict[self._clo_node]['next_nodes'].append(uid)
+        dag_dict[pipe._clo_node]['next_nodes'].append(uid)
+
+        config = self._concat_config(self._config, pipe._config)
+        return Pipeline(dag_dict, uid, config)
```
junjiejiangjjj, line 146: Don't return new pipeline, just return self

tests/unittests/runtime/test_user_pipeline.py
```diff
@@ -0,0 +1,145 @@
...
+    def test_registry(self):
+        pipe1 = (Pipeline.input('a', 'b')
+                 .map('a', 'c', ops.add_operator(10))
+                 .map(('b', 'c'), 'd', ops.sub_operator())
+                 .output('c', 'd'))
+        c1, d1 = pipe1(1, 2).get()
+        self.assertEqual(c1, 11)
+        self.assertEqual(d1, -9)
+
+    def test_multi_pipe(self):
```
junjiejiangjjj, line 94:
```python
p1 = pipe.input('a')
p2 = p1.map(...)
p3 = p1.map(...)

p2.output(...)
p3.output(...)

it seems p2 and p3 have the same dag, can it work correctly?

Comment (shiyu22) [reply tests/unittests/runtime/test_user_pipeline.py:94]: It works.
```bash
p1 = Pipeline.input('a')
p2 = p1.map('a', 'c', ops.add_operator(10))
p3 = p1.map('a', 'c', ops.add_operator(10))

res1 = p2.output('c')(1)
res2 = p3.output('c')(2)
print(res1.get(), res2.get())
```

PR #2232, opened 2022-10-27
Add pipeline test

junjiejiangjjj:

### DIFF
--- a/towhee/runtime/data_queue.py
+++ b/towhee/runtime/data_queue.py
@@ -24,7 +24,7 @@ class DataQueue:
     Col-based storage.
     """
 
-    def __init__(self, schema_info, max_size=0):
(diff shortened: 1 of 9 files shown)


towhee/runtime/data_queue.py
```diff
@@ -130,6 +130,13 @@ def get_dict(self) -> Optional[Dict]:
             ret[names[i]] = data[i]
         return ret
 
+    def reset_max_size(self, max_size):
```
GuoRentong, line 133: consider this:
implement this method as a property setter

GuoRentong (changes requested):

towhee/runtime/data_queue.py
junjiejiangjjj: done

PR #2241, opened 2022-10-31
Updata pipeline and add examples

shiyu22:

### DIFF
--- a/towhee/runtime/operator_manager/operator_action.py
+++ b/towhee/runtime/operator_manager/operator_action.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pickle
-import marshal
-
 
 # pylint: disable=protected-access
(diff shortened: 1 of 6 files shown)


towhee/runtime/operator_manager/operator_action.py
```diff
@@ -92,6 +96,20 @@ def serialize(self):
...
+            return {
+                'operator': self._loaded_fn,
+                'type': self._type,
+                'init_args': None,
+                'init_kws': None,
+                'tag': None
+            }
         elif self._type == 'callable':
-            raise ValueError('Callable not supported yet.')
+            self._loaded_fn = pickle.dumps(self._fn)
```
junjiejiangjjj, line 108: Can the dag dump into a JSON file? if not, in the local model, we just store the func itself

shiyu22: Done, return the func.

PR #2260, opened 2022-11-04
add upload s3 function

zhujiming:


towhee/command/s3.py
```diff
@@ -0,0 +1,177 @@
...
+
+from towhee.utils.thirdparty.boto3_utils import boto3
+import re
+import os
+import math
+import hashlib
+from boto3.s3.transfer import TransferConfig
+from botocore.exceptions import ClientError
+
+S3_CONF = {
```
shiyu22, line 23: Is there any risk in exposing these keys?

Issue #2271, opened 2022-11-10
[DesignProposal]: Perfermence Tracer for Nodes


shiyu22:
### Background and Motivation

Each pipeline contains several nodes and we need support to trace the performance of each node. And the metric of performance includes `preload time`, `runtime`, and `queue time` for the node, and `avg`, `max`, and `min` for the pipeline(total time), for example we will show the following performance results:
```
nodes number: 2
 |       node        | count | success |  qps  | preload_avg |   queue_avg  |  run_avg   |
 |       func1       |   xx  |    xx   |  xxx  |      xxx    |     xxx      |     xxx    |
 |       func2       |   xx  |    xx   |  xxx  |      xxx    |     xxx      |     xxx    |

pipeline
|   avg     |   max     |    min     |
|   xxx     |    xxx    |    xxx     |
```

### Design

Set tracer for the pipeline or the specific node to config the trace. 

For example, if we declare a pipeline with `.output(tracer=True)`, it will trace the performance for all the nodes except input and output. After we run the pipeline, we can run `pipeline.performance_tracer()` to show the performance.

```python
from towhee.dc2 import pipe, ops

insert_pipe = ( pipe.input('path')
              .map('path', 'img', ops.image_decode())
              .map('img', 'vectors', ops.image_embedding.timm(model_name='resnet50'))
              .map(('vectors', 'path'), 'milvus_res', ops.ann_insert.milvus(uri=f'tcp://127.0.0.1:19530/collection')) 
              .output('milvus_res', tracer=True)
     )

for p in paths:
    insert_pipe(p)

insert_pipeline.performance_tracer()
```

And we can also just set tracer for the specific node, such as we can trace the **ops.image_embedding.timm** node with `config={'tracer': True}`.

```python
insert_pipe = ( pipe.input('path')
              .map('path', 'img', ops.image_decode())
              .map('img', 'vectors', ops.image_embedding.timm(model_name='resnet50'), config={'tracer': True})
              .map(('vectors', 'path'), 'milvus_res', ops.ann_insert.milvus(uri=f'tcp://127.0.0.1:19530/collection')) 
              .output('milvus_res')
     )

for p in paths:
    insert_pipe(p)

insert_pipeline.performance_tracer()
```

PR #2273, opened 2022-11-11
Add performence tracer for nodes

shiyu22:
#2271

### DIFF
--- a/towhee/runtime/performance_profiler.py
+++ b/towhee/runtime/performance_profiler.py
@@ -62,8 +62,8 @@ class PipelineProfiler:
 
     def show(self):
         print('Input: ', self.data)
-        print('Total time:', round(self.time_out - self.time_in, 3))
-        headers = ['node', 'ncalls', 'total_time', 'init', 'wait_data', 'call_op', ' output_data']
+        print('Total time(s):', round(self.time_out - self.time_in, 3))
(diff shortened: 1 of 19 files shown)


towhee/runtime/runtime_pipeline.py
```diff
@@ -36,11 +40,13 @@ def __init__(self,
                  nodes: Dict[str, NodeRepr],
                  edges: Dict[str, Any],
                  operator_pool: 'OperatorPool',
-                 thread_pool: 'ThreadPoolExecutor'):
+                 thread_pool: 'ThreadPoolExecutor',
+                 time_profiler: 'TimeProfiler' = TimeProfiler()):
```
junjiejiangjjj, line 44: Use None as the default value, and in `__init__`,  create TimeProfiler if the input is None.

towhee/runtime/performance_profiler.py
```diff
@@ -0,0 +1,245 @@
...
+
+class TimeProfiler:
+    """
+    TimeProfiler to record the timestamp.
+    """
+    def __init__(self):
+        self._enable = False
+        self.time_record = []
+
+    def record(self, name, event, pipe_input=None):
```
junjiejiangjjj, line 232: input & output nodes also need tracer.

shiyu22: Get it, and do we need to show the performance for them?

junjiejiangjjj: yeah

PR #2310, opened 2022-11-29
Add model to triton with onnx

shiyu22:


tests/unittests/triton/test_model_to_triton.py
```diff
@@ -0,0 +1,119 @@
...
+import unittest
+from pathlib import Path
+
+import torch
+import torchvision
+from towhee.runtime.node_config import NodeConfig
+from towhee.serve.triton.model_to_triton import ModelToTriton
+
+
+public_path = Path(__file__).parent.parent.resolve()
```
junjiejiangjjj, line 24: `from tempfile import TemporaryDirectory `
Use this lib to create a temp dir, it makes sure that temp dir will be deleted

PR #2351, opened 2022-12-21
Solve op_loader loading unexpected cognominal module issue

Chiiizzzy:

### DIFF
--- a/tests/unittests/engine/test_operator_loader.py
+++ b/tests/unittests/engine/test_operator_loader.py
@@ -43,3 +43,9 @@ class TestOperatorLoader(unittest.TestCase):
         op = ops.local.old_format_op_inconsistent(10)
         with self.assertRaises(FileNotFoundError):
             op(10)
+
+    def test_op_name(self):
+        op1 = ops.local.cal_sum(10)
+        self.assertEqual(op1(10), 20)
+        op2 = ops.local.cal_diff(10)
+        self.assertEqual(op2(10), 0)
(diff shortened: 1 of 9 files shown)


towhee/runtime/operator_manager/operator_loader.py
```diff
@@ -110,8 +110,9 @@ def load_operator_from_path(self, path: Union[str, Path], arg: List[Any], kws: D
                 The `Operator` output.
         """
         path = Path(path)
-        fname = Path(path).stem
-        modname = 'towhee.operator.' + fname
+        fname = path.stem
+        op_name = path.parent.parent.parent.stem
```
junjiejiangjjj, line 114: Why not use the `function`  param

towhee/engine/operator_loader.py
```diff
@@ -124,8 +124,9 @@ def load_operator_from_path(self, path: Union[str, Path], arg: List[Any], kws: D
                 The `Operator` output.
         """
         path = Path(path)
-        fname = Path(path).stem
-        modname = 'towhee.operator.' + fname
+        fname = path.stem
+        op_name = function.split('/')[0].replace('-', '_')
+        modname = 'towhee.operator.' + op_name + '.' + fname
```
junjiejiangjjj, line 129: It seems modname can not be unique.

tests/unittests/engine/test_operator_loader.py
```diff
@@ -43,3 +43,7 @@ def test_inconsistent(self):
         op = ops.local.old_format_op_inconsistent(10)
         with self.assertRaises(FileNotFoundError):
             op(10)
+
+    def test_op_name(self):
```
junjiejiangjjj, line 47: We need to use two ops with the same dir struct to test the case

PR #2360, opened 2022-12-29
Add AutoConfig for pipeline

shiyu22:


towhee/runtime/node_config.py
```diff
@@ -178,3 +178,161 @@ def from_dict(triton_info: Dict[str, Any]):
...
+        Examples:
+            >>> from towhee.dc2 import pipe, AutoConfig
+            >>> p = (pipe.input('a')
+            ...          .flat_map('a', 'b', lambda x: [y for y in x], config=AutoConfig.LocalCPUConfig())
+            ...          .output('b'))
+        """
+        device_info = {
+            'device': -1
+        }
+        return [device_info]
```
junjiejiangjjj, line 207: Return a config object is better than a list or dict

shiyu22: Yeah, and add TowheeConfig for mapping.

PR #2432, opened 2023-02-23
Use new hub interface to download op

junjiejiangjjj:


towhee/hub/downloader.py
```diff
@@ -0,0 +1,178 @@
...
+                if os.path.exists(f.name):
+                    os.remove(f.name)
+
+    def download(self):
+        futures = []
+        with ThreadPoolExecutor(max_workers=5) as pool:
+            for local_file, url in self._hub_files.local_url_pair():
+                if not local_file.is_file():
+                    futures.append(pool.submit(_Downloader.download_url_to_file, url, local_file))
+            _ = [i.result() for i in futures]
```
xiaofan-luan, line 158: exception handling?

xiaofan-luan:
overall lgtm

towhee/hub/downloader.py
junjiejiangjjj: This function would raise the exception, and the `operator_loader` will catch the exception, print logs and stop running.

PR #2462, opened 2023-03-09
Rebuild dag with pipelines

shiyu22:


towhee/runtime/pipeline.py
```diff
@@ -16,9 +16,11 @@
 import uuid
 from copy import deepcopy
 
+from towhee.operator.nop import NOPNodeOperator
```
junjiejiangjjj, line 19: Do not use Operator directly, all ops should load by `operator_pool`

towhee/runtime/pipeline.py
```diff
@@ -138,6 +140,11 @@ def map(self, input_schema, output_schema, fn, config=None) -> 'Pipeline':
         output_schema = self._check_schema(output_schema)
         input_schema = self._check_schema(input_schema)
 
+        if isinstance(fn, RuntimePipeline):
```
junjiejiangjjj, line 143: We may put the rebuild into `dag_repr`

tests/unittests/triton/test_pipeline_builder.py
```diff
@@ -95,6 +95,13 @@ def test_failed(self):
                     'type': 'map',
                     'param': None
                 },
+                'op_info': {
+                    'operator': 'NOPNodeOperator',
+                    'type': 'built_in',
```
junjiejiangjjj, line 100: The type is the node's type,  nothing with the ops.

PR #2469, opened 2023-03-14
Add data visualizer for pipeline

Chiiizzzy:


towhee/tools/data_visualizer.py
```diff
@@ -0,0 +1,120 @@
...
+    def show(self, tablefmt=None, show_type=True):
+        for k in self._node_collections.keys():
+            NodeVisualizer(k, self._node_collections[k]).show(tablefmt=tablefmt, show_type=show_type)
+            print()
+
+    @property
+    def nodes(self):
+        return list(self._node_collections.keys())
+
+    def __getitem__(self, name: str):
```
junjiejiangjjj, line 119: Show all nodes' names,  if the user gives an error node name

PR #2476, opened 2023-03-20
Rename the node with pipeline

shiyu22:


towhee/runtime/operator_manager/operator_action.py
```diff
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+from towhee.runtime.dag_repr import DAGRepr
```
junjiejiangjjj, line 15: `DAGRepr`   <--> `OperatorAction`  we'd better remove the circular reference.

junjiejiangjjj, line 15: Can you remove the dependency of DAGRepr.

shiyu22: Done.

PR #2480, opened 2023-03-21
Add revision and get_latest for operator

shiyu22:


tests/unittests/runtime/test_user_pipeline.py
```diff
@@ -406,3 +406,13 @@ def test_flatmap_coverage(self):
         )
         res = p(4).to_list()
         self.assertEqual(res, [[60], [40]])
+
+    def test_revision_latest(self):
+        p = (
+            Pipeline.input('a')
+            .map('a', 'b', ops.test_revision().revision('v1').get_latest())
```
junjiejiangjjj, line 413: .latest()  is enough

shiyu22: But it duplicated with the `latest` attr code:
```python
    @property
    def latest(self):
        return self._latest
```

junjiejiangjjj: > But it duplicated with the `latest` attr code:
> 
> ```python
>     @property
>     def latest(self):
>         return self._latest
> ```

is_latest ?

PR #2481, opened 2023-03-22
Fix docker build using cache issue

Chiiizzzy:


towhee/serve/triton/dockerfiles/DockerfileCuda117dev
```diff
@@ -30,4 +24,7 @@ COPY ./dag.pickle /workspace/dag.pickle
 
 COPY ./server_config.json /workspace/server_config.json
 
+RUN pip install --no-cache-dir -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple towhee --pre
```
shiyu22, line 27: I think it only works the first time to install the towhee because you changed the dockerfile. And force build maybe need to [update the docker build command](https://stackoverflow.com/questions/35594987/how-to-force-docker-for-a-clean-build-of-an-image).

PR #2486, opened 2023-03-23
Update op modname with tag

shiyu22:


tests/unittests/runtime/test_operator_loader.py
```diff
@@ -52,3 +54,15 @@ def test_op_name(self):
...
         self.assertEqual(op2(10), 0)
+
+    def test_load_operator_from_path(self):
+        op1 = ops.test_revision()
+        self.assertEqual(op1()[0], 'main')
+
+        op1 = ops.test_revision().revision('no-init')
+        self.assertEqual(op1()[0], 'no-init')
+
+        operator_cache = Path(__file__).parent.parent.resolve() / 'mock_operators'
```
junjiejiangjjj, line 65: from .. import operator_cache

PR #2497, opened 2023-03-29
Add reduce node

junjiejiangjjj:


towhee/runtime/dag_repr.py
```diff
@@ -420,6 +423,9 @@ def rebuild_dag(dag, sub_uid, op_info, iter_info, input_schema, output_schema):
         elif iter_type == WindowAllConst.name:
             new_dag[sub_uid]['iter_info'] = {'type': WindowAllConst.name,
                                              'param': None}
+        elif iter_type == ReduceConst.name:
+            new_dag[sub_uid]['iter_info'] = {'type': ReduceConst.name,
```
shiyu22, line 427: I think it's right, and can you add UT in test_dag_pipeline?

junjiejiangjjj: fixed

PR #2501, opened 2023-04-03
Add tracer in debug

Chiiizzzy:


towhee/runtime/runtime_pipeline.py
```diff
@@ -49,23 +50,31 @@ def __init__(self,
                  edges: Dict[str, Any],
                  operator_pool: 'OperatorPool',
                  thread_pool: 'ThreadPoolExecutor',
-                 enable_profiler: bool = False):
+                 enable_profiler: bool = False,
+                 enable_tracer: bool = False,
```
junjiejiangjjj, line 54: `enable_tracer` can be removed, we could determine whether `trace_edges`  is empty or not.

Issue #2674, opened 2023-11-20
[Bug]: application cannot start. `from towhee import AutoPipes` crashes


dove-young:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

application cannot start. `from towhee import AutoPipes` crashes

```
 python towhee-test.py
Traceback (most recent call last):
  File "/root/jupyter/instana-slack-app-dev/towhee-test.py", line 1, in <module>
    from towhee import AutoPipes
  File "/root/jupyter/virtualenv/slack-dev-3.10/lib/python3.10/site-packages/towhee/__init__.py", line 20, in <module>
    from towhee.runtime import register, pipe, ops, accelerate, AutoConfig, AutoPipes
  File "/root/jupyter/virtualenv/slack-dev-3.10/lib/python3.10/site-packages/towhee/runtime/__init__.py", line 17, in <module>
    from .pipeline import Pipeline as pipe
  File "/root/jupyter/virtualenv/slack-dev-3.10/lib/python3.10/site-packages/towhee/runtime/pipeline.py", line 18, in <module>
    from towhee.runtime.check_utils import TupleForm
  File "/root/jupyter/virtualenv/slack-dev-3.10/lib/python3.10/site-packages/towhee/runtime/check_utils.py", line 36, in <module>
    class TupleForm(BaseModel):
  File "/root/jupyter/virtualenv/slack-dev-3.10/lib/python3.10/site-packages/towhee/runtime/check_utils.py", line 38, in TupleForm
    schema_data: Optional[Tuple[constr(regex='^[a-z][a-z0-9_]*$'), ...]]
TypeError: constr() got an unexpected keyword argument 'regex'
```

### Expected Behavior

application runs

### Steps To Reproduce

```markdown
1. in Python 3.10 env
2. `pip install towhee pydantic_core`
3. run a python script with `from towhee import AutoPipes`
4. see error


pip list | grep pydantic
pydantic                 1.10.13
pydantic_core            2.14.4
```
```


### Environment

```markdown
- Towhee version(e.g. v0.1.3 or 8b23a93): 1.1.2
- OS(Ubuntu or CentOS): RHEL 8
- CPU/Memory: 8 CPU 32G Memory
- GPU:
- Others: Python 3.10
```

### Anything else?

And actually after install with `pip install towhee towhee.models`, it does not run. 
```
Traceback (most recent call last):
[...]

PR #2675, opened 2023-11-21
fix issue 2674 - enable towhee to work with pydantic 2.5.x

dove-young:
```
pydantic                 2.5.1
pydantic_core            2.14.3
```

Fix issue https://github.com/towhee-io/towhee/issues/2674

Fix issue https://github.com/towhee-io/towhee/issues/2676


junjiejiangjjj:
Very tks for this pr, however, this change does not work in environments with pydantic<2. If you want to modify it, you need to make sure it works in both pydantic 1.x and 2.x.

